Issue #11, opened 2019-08-14
Use latest go version


mastertinner:
The default Go version this action uses is quite outdated. Having to set the version manually is rather tedious. Is there a way to just use the latest stable version? Also, if I set it to `go-version: "1.12"`, I would expect it to use the latest `1.12.x` (currently `1.12.8`) but instead it uses `1.12.0`.

subosito:
@mastertinner  I've made a PR for this issue. Once reviewed and merged, you can use `1.12.x` which resolves to `1.12.8`.

damccorm:
Should be fixed now that @subosito's PR has merged

PR #13, opened 2019-08-16
Handle .x version syntax with latest release

subosito:
This PR adds the ability to resolve .x version syntax to the latest release, e.g.: `1.10.x` as `1.10.8`.

This PR is a solution for https://github.com/actions/setup-go/issues/11.


src/installer.ts
```diff
@@ -123,8 +127,70 @@ function normalizeVersion(version: string): string {
...
+    }
+
     //append patch version if not available
     return version.concat('.0');
   }
   return version;
 }
+
+async function determineVersion(version: string): Promise<string> {
+  if (!version.endsWith('.x')) {
```
damccorm: Could we do something similar to `normalizeVersion` here, but instead of appending 0s at the end, append a `.x`? What I'm after is, I want people to just be able to specify `1.10` and have that resolve to `1.10.x`, since referring to versions like that is pretty common.

src/installer.ts
```diff
@@ -123,8 +127,70 @@ function normalizeVersion(version: string): string {
...
+  // clean .x syntax: 1.10.x -> 1.10
+  const trimmedVersion = version.slice(0, version.length - 2);
+
+  const versions = await getPossibleVersions(trimmedVersion);
+
+  core.debug(`evaluating ${versions.length} versions`);
+
+  if (version.length === 0) {
+    core.debug('match not found');
+    return trimmedVersion;
```
damccorm: If no matching versions are found should we just throw? Are there any cases where we don't match a version here, but we succeed later? We might get a clearer error that way

src/installer.ts
```diff
@@ -123,8 +127,70 @@ function normalizeVersion(version: string): string {
...
+  return tags
+    .filter(tag => tag.ref.match(/go\d+\.[\w\.]+/g))
+    .map(tag => tag.ref.replace('refs/tags/go', ''));
+}
+
+async function getPossibleVersions(version: string): Promise<string[]> {
+  const versions = await getAvailableVersions();
+  const possibleVersions = versions.filter(v => v.startsWith(version));
+
+  const versionMap = new Map();
```
damccorm: Rather than creating a map here, can you just do something like `return possibleVersions.sort()` (an alphabetic sort which matches the semver versioning ordering). I think what you're doing here is basically just doing that.

package-lock.json
```diff
@@ -5,26 +5,26 @@
   "requires": true,
   "dependencies": {
     "@actions/core": {
-      "version": "file:toolkit/actions-core-0.0.0.tgz",
+      "version": "file:E:/github/setup-go/toolkit/actions-core-0.0.0.tgz",
```
damccorm: Looks like these file paths got made into absolute paths, could you change them back to relative? We really just need to consume the toolkit from npm, just haven't gotten around to changing that yet.

damccorm: EDIT - updated to consume from npm so this no longer applies, that's why you have conflicts now though

src/installer.ts
subosito: What I am doing here is sorting using semver, which different with regular sort function. 

`sort` returns:

```
[ '1.10',
  '1.10.1',
  '1.10.2',
  '1.10.3',
  '1.10.4',
  '1.10.5',
  '1.10.6',
  '1.10.7',
  '1.10.8',
  '1.10beta1',
  '1.10beta2',
  '1.10rc1',
  '1.10rc2' ]
```

While this returns:

```
[ '1.10.8',
  '1.10.7',
  '1.10.6',
  '1.10.5',
  '1.10.4',
  '1.10.3',
  '1.10.2',
  '1.10.1',
  '1.10',
  '1.10rc2',
  '1.10rc1',
  '1.10beta2',
  '1.10beta1' ]
```

src/installer.ts
subosito: Thanks for the suggestion, I've pushed an update to handle this.

src/installer.ts
damccorm: Oh good point, sorry my mistake

damccorm:
This looks great - thanks for doing this!

src/installer.ts
```diff
@@ -122,9 +129,87 @@ function normalizeVersion(version: string): string {
...
+
+async function getLatestVersion(version: string): Promise<string> {
+  // clean .x syntax: 1.10.x -> 1.10
+  const trimmedVersion = version.slice(0, version.length - 2);
+
+  const versions = await getPossibleVersions(trimmedVersion);
+
+  core.debug(`evaluating ${versions.length} versions`);
+
+  if (version.length === 0) {
```
tie: `versions.length`.

Issue #14, opened 2019-08-16
Add $GOPATH/bin to $PATH


lmittmann:
When installing tools via `go get` (e.g. golint) they can not be ran without adding `$GOPATH/bin` to `$PATH` beforehand.

So for using e.g. golint one would need to write a step:
```yml
- run: |
    export PATH=$PATH:$(go env GOPATH)/bin
    go get -u golang.org/x/lint/golint
    golint -set_exit_status ./...
```

It would be cool, if setup-go could already do the `export PATH=$PATH:$(go env GOPATH)/bin` part, as this is a common need.

geoah:
GOBIN would be a better choice I think for this

lmittmann:
`$GOBIN` could be used as well, but it also needs to be added to `$PATH`. Since `go env GOBIN` is not set and `$GOPATH/bin` is the default installation path it's just simpler to add the default path to `$PATH`.

damccorm:
I think this should probably be resolved together with #12. Given [my most recent comment](https://github.com/actions/setup-go/issues/12#issuecomment-522593317) in that issue, are you ok with closing here and following up there?

lmittmann:
@damccorm Let's track this issue here, as it seems like #12 is quite fuzzy.

mvdan:
I think you're overthinking this. The official Go docker images simply hard-code `PATH`, as you can see here: https://github.com/docker-library/golang/blob/3a6407a6ff134ef6a0364ac061b0808f990ea14e/1.13-rc/alpine3.10/Dockerfile#L59-L60

Why not do the same? If anyone wants to set up a custom `GOPATH` or `GOBIN`, they can set up a custom `PATH` too. I'd prefer `PATH` to be simple and predictable.

leighmcculloch:
You can avoid this issue by instead using the actions `container` param to run all the commands in the official `golang` Docker image. You can see what that looks like here:
https://github.com/stellar/go-xdr/blob/3aa3546/.github/workflows/build.yml

mvdan:
@leighmcculloch I assume that won't work for testing on mac and windows, though, so fixing this bug is still very useful.

sanemat:
I prefer this solution below. This works on windows, ubuntu, and mac.
```
    steps:
    - name: setup go
      uses: actions/setup-go@v1
      with:
        go-version: 1.x
    - name: setup env
      run: |
        echo "::set-env name=GOPATH::$(go env GOPATH)"
[...]

Issue #12, opened 2019-08-15
Setting GO environment variables (GOPATH, GOBIN)


radu-matei:
Ref: https://github.com/actions/setup-go/blob/818723587f041a921450ba8adf4e69063a18d821/src/installer.ts#L105-L115

```
2019-08-15T22:05:36.9141345Z env:
2019-08-15T22:05:36.9141441Z   GOROOT: /opt/hostedtoolcache/go/1.12.0/x64
2019-08-15T22:05:36.9141482Z ##[endgroup]
2019-08-15T22:05:37.0161150Z LEIN_HOME=/usr/local/lib/lein
2019-08-15T22:05:37.0161985Z M2_HOME=/usr/share/apache-maven-3.6.1
2019-08-15T22:05:37.0162146Z BOOST_ROOT=/usr/local/share/boost/1.69.0
2019-08-15T22:05:37.0162236Z GOROOT_1_11_X64=/usr/local/go1.11
2019-08-15T22:05:37.0162330Z ANDROID_HOME=/usr/local/lib/android/sdk
2019-08-15T22:05:37.0162641Z JAVA_HOME_11_X64=/usr/lib/jvm/zulu-11-azure-amd64
2019-08-15T22:05:37.0162818Z LANG=C.UTF-8
2019-08-15T22:05:37.0162931Z INVOCATION_ID=e226194a393e43d59aba6bd4c85af030
2019-08-15T22:05:37.0163149Z JAVA_HOME_12_X64=/usr/lib/jvm/zulu-12-azure-amd64
2019-08-15T22:05:37.0163325Z ANDROID_SDK_ROOT=/usr/local/lib/android/sdk
2019-08-15T22:05:37.0163420Z RUNNER_TOOL_CACHE=/opt/hostedtoolcache
2019-08-15T22:05:37.0163674Z JAVA_HOME=/usr/lib/jvm/zulu-8-azure-amd64
2019-08-15T22:05:37.0163967Z RUNNER_TRACKING_ID=github_28b0e254-504e-46ce-b1ab-b1db282d32b4
2019-08-15T22:05:37.0164043Z GITHUB_ACTIONS=true
2019-08-15T22:05:37.0164178Z DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
2019-08-15T22:05:37.0164238Z USER=runner
2019-08-15T22:05:37.0164303Z GITHUB_HEAD_REF=
[...]

PR #19, opened 2019-09-03
Set GOPATH to the default directory if unset

crazy-max:
This PR should fix issues linked to GOPATH definition (#12 #14).
TravisCI also has pretty much [this behavior](https://github.com/travis-ci/gimme/blob/master/Dockerfile#L13-L14) while initializing Go with [gimme](https://github.com/travis-ci/gimme) and the official Go Docker image [too](https://github.com/docker-library/golang/blob/3a6407a6ff134ef6a0364ac061b0808f990ea14e/1.13-rc/alpine3.10/Dockerfile#L59-L60).


kjk:
FYI: looks like you've inadvertently changed the line endings of the files, so unless the diff is told to ignore whitespace changes it's hard to tell what changed.

crazy-max:
@kjk Yeah looks like tsc fucked up ^^
Scroll down to TS definition to look at my changes

crazy-max:
@damccorm Some feedback on this ?

bryanmacfarlane:
Feedback:
- the diff needs to be cleaned up so it's just what's changed.  You may want to re-fork or something.  I just did a massive rewrite and master is v2-beta and v1 is in releases/v1. If we do this, then v2 only.
- I don't think this is the right thing to do for everyone.  I think there should be a bool input whether to setGoPath and it should default to false. There's post 1.13 scenarios with modules where you absolutely don't want to do this and those scenarios are becoming more and more common as folks adopt modules which must have your code outside of GOPATH.
- because of points above, I wonder whether we should do this.  As time goes on, the feature is less needed.

bryanmacfarlane:
also, do an `npm run pre-checkin` which will reformat.  Set line endings to auto.  are you coding on windows?

crazy-max:
@bryanmacfarlane Yeah I'm currently on Windows but will move back to Linux in few hours ;)

bryanmacfarlane:
I might also set some repo level git line level ending settings ...

PR #30, opened 2019-11-21
Change getFileName to do architecture lookups and add arm support

hross:
Added arm support to go setup and refactored architecture check.


dakale:
LGTM but i dont have write access either

PR #32, opened 2019-12-14
Add support for major release selection

### LINKED ISSUE #21: Support tip?
I looked through the [installer.js](https://github.com/actions/setup-go/blob/master/lib/installer.js) script and it looked to me like only tags from https://github.com/golang/go/tags are supported as valid versions to install. 

I would imagine that some users would also want to test their Go programs against `tip` to ensure that upcoming language changes/additions don't break. 

Go even provides a way to install `tip` via https://godoc.org/golang.org/dl/gotip (once you have a single Go version installed that is).

What do you think?

### LINKED ISSUE #31: Use the latest release
#11 gave us the ability to use the latest version for a specific release, in other other words - to care about major and minor, but not a patch version.

Is it possible to specify the latest minor as well? Just a `1.x` which will take the latest Go release, today it will be `1.13.4`, but after the release of `1.14` (somewhere in February) `1.x` become `1.14.x`.

Basically, the same as Travis does. Thank you.

tie:
This PR adds 'latest' version alias that resolves to the latest available release.

The `getPossibleVersions` function filters using `startsWith(version)`, so calling `getLatestVersion(version)` with an empty string will return the latest release.

Closes #31.

### DIFF
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -167,6 +167,10 @@ function normalizeVersion(version: string): string {
 }
 
 async function determineVersion(version: string): Promise<string> {
+  if (version == 'latest') {
+    return await getLatestVersion('');
+  }
+
   if (!version.endsWith('.x')) {
     const versionPart = version.split('.');
 
@@ -188,7 +192,7 @@ async function getLatestVersion(version: string): Promise<string> {
 
   core.debug(`evaluating ${versions.length} versions`);
 
-  if (version.length === 0) {
+  if (versions.length === 0) {
     throw new Error('unable to get latest version');
   }
 
--- a/__tests__/installer.test.ts
+++ b/__tests__/installer.test.ts
@@ -89,6 +89,18 @@ describe('installer tests', () => {
         expect(fs.existsSync(path.join(goDir, 'bin', 'go'))).toBe(true);
       }
     }, 100000);
+
+    it('Acquires latest release version of go if using latest and matching version is installed', async () => {
+      await installer.getGo('latest');
+      const goDir = path.join(toolDir, 'go', '1.13.0', os.arch());
+
+      expect(fs.existsSync(`${goDir}.complete`)).toBe(true);
+      if (IS_WINDOWS) {
+        expect(fs.existsSync(path.join(goDir, 'bin', 'go.exe'))).toBe(true);
+      } else {
+        expect(fs.existsSync(path.join(goDir, 'bin', 'go'))).toBe(true);
+      }
+    }, 100000);
   });
 
   it('Throws if no location contains correct go version', async () => {


src/installer.ts
```diff
@@ -167,6 +167,10 @@ function normalizeVersion(version: string): string {
 }
 
 async function determineVersion(version: string): Promise<string> {
+  if (version == 'latest' || version == 'tip') {
```
zikaeroh: I left this comment on the commit directly, but I would strongly advise against checking tip here. If it doesn't actually install tip, then it's going to be misleading when someone ends up with the latest stable build instead.

#21 is about actually building tip, like `gimme` does on Travis. Needing an extra step in the workflow anyway makes this check not worthwhile, since the step would have worked without this change (see https://github.com/actions/setup-go/issues/21#issuecomment-565704236).

tie: I left the comment on the commit too. https://github.com/tie/setup-go/commit/ff6bc1640078e249753d4f5ff3bb74eb647df2a6#commitcomment-36429256

Why would the following (https://github.com/actions/setup-go/issues/21#issuecomment-565704236) work when `tip` is in a build matrix? Shouldn’t it be ignored because of the `matrix.go != 'tip'`?

```yaml
- name: Install Go
  if: matrix.go != 'tip'
  uses: actions/setup-go@v1
  with:
    go-version: ${{ matrix.go }}
```

zikaeroh: Matrixes are separate from actions. If it's not tip, `setup-go` runs. If it's tip, then `setup-go` does not run, and `Go` is installed via `gotip` using the builder's preinstalled Go version: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners

It doesn't matter what version you start with to build tip; Go has no problem building itself from versions even as old as Go 1.4 (the last version written in C).

I plan to update my comment to make it more efficient; installing `gotip` is just overhead that's wasting time.

tie: Ah, sorry, didn’t read the whole comment.

>No need to run setup-go to bootstrap this stage, since the builders come with a Go version already installed.

zikaeroh: Of course, removing that assumption would be preferred; I don't know if self-hosted builders can make all of the same assumptions as Github-run ones that bundle a bunch of stuff. For tip support in setup-go, I'd probably remove that assumption (and it won't be too slow, thanks to the caching that's happening when setup-go runs).

tie: In this case there is really no need for `tip` alias.

tie: I’ll look into adding built-in `tip` support in a separate PR later.

zikaeroh:
Thanks for removing the `tip` stuff. Can you modify your PR message now that `tip` isn't included to better reflect the change? I wouldn't want this to be merged and close the wrong issue (or mention things the PR no longer does).

Another suggestion would be to allow an unset `go-version` to imply `latest` (for lazy users who don't care), but I'm not sure how good of an idea that is when the builders already include some version of Go (and it's better for build pipelines to be consistent-ish, where picking `1.13.x` would be better than saying nothing).

tie:
Updated PR message to reflect the change.

tie:
Update commit message too.

bryanmacfarlane:
With the new v2-beta rewrite, should this just be 1.* . (that supports semver ranges)

bryanmacfarlane:
See comment ☝️

PR #34, opened 2019-12-15
Add support for gotip and latest

### LINKED ISSUE #21: Support tip?

### LINKED ISSUE #31: Use the latest release

### LINKED ISSUE #35: Broken version normalization
PR #13 broke the way version normalization works. Since it’s merged, now `1.10` resolves to `1.10.8` or whatever the latest version is instead of `1.10.0`. This is *not* the behavior Go user expects.

tie:
This PR adds support for building Go `tip` from source and an alias `latest` for the major release.

Closes #21
Closes #31
Closes #32
Closes #35

### DIFF
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -99,7 +99,9 @@ async function acquireGo(version: string, gotipRef: string, bootstrapGo: string)
         if (!workTree) {
           // We found gotip.git in cache, but not the work tree.
           //
-          workTree = path.join(extPath, filename);
+          // Also append commit hash to prevent conflicts with
+          // other checkouts.
+          workTree = path.join(extPath, filename, commitHash);
           // Work tree must exist, otherwise Git will complain
           // that “this operation must be run in a work tree”.
           await io.mkdirP(workTree);
@@ -122,6 +124,10 @@ async function acquireGo(version: string, gotipRef: string, bootstrapGo: string)
         const env = {
           'GOROOT_BOOTSTRAP': bootstrap,
           ...process.env,
+          // Tell Git where to look for the repo. Git will be invoked
+          // by cmd/dist to find Go version and write VERSION file.
+          'GIT_DIR': gitDir,
+          'GIT_WORK_TREE': workTree,
           // Note that while we disable Cgo for tip builds, it does
           // not disable Cgo entirely. Moreover, we override this
           // value in setGoEnvironmentVariables with whatever the
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -112,8 +112,9 @@ async function acquireGo(version: string, gotipRef: string, bootstrapGo: string)
 
         // The make.bat script on Windows is not smart enough
         // to figure out the path to bootstrap Go toolchain.
-        // Make script will show descriptive error message even
-        // if we don’t find Go installation on the host.
+        // We have to spoonfeed it with GOROOT_BOOTSTRAP value.
+        // At least make script will show descriptive error message
+        // even if we don’t find Go installation on the host.
         let bootstrap: string = '';
         if (bootstrapGo) {
           bootstrap = await executil.goEnv('GOROOT', bootstrapGo);
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -186,7 +186,7 @@ function getFileName(version: string): string {
   const arch: string = arches[osArch] || arches['default'];
   let ext: string;
   if (version == 'tip') {
-    // Git work tree for tip builds does not have an externsion.
+    // Git work tree for tip builds does not have an extension.
     ext = '';
   } else if (osPlat == 'win32') {
     ext = '.zip';
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -63,8 +63,7 @@ async function acquireGo(version: string, gotipRef: string, bootstrapGo: string)
       let workTree: string;
       let commitHash: string;
       // Avoid cloning multiple times by caching git dir.
-      // Empty string means that we don’t care about arch.
-      gitDir = tc.find('gotip', 'master', '')
+      gitDir = tc.find('gotip', '0.0.0-devel', 'noarch');
       if (!gitDir) {
         gitDir = path.join(extPath, 'gotip.git');
         workTree = path.join(extPath, filename);
@@ -74,6 +73,13 @@ async function acquireGo(version: string, gotipRef: string, bootstrapGo: string)
 
         // Extract current commit hash.
         commitHash = await executil.gitRevParse(gitDir, 'HEAD');
+
+        // Add cache for git dir. Note that in the current tool-cache
+        // implementation adding result of find to the cache actually
+        // purges both. That is, we can’t update the cache explicitly
+        // and tool-cache assumes we won’t change tool in the cache.
+        // And in the other branch we break that assumption.
+        gitDir = await tc.cacheDir(gitDir, 'gotip', '0.0.0-devel', 'noarch');
       } else {
         // We don’t have a work tree (yet) in this case.
         workTree = '';
@@ -84,11 +90,9 @@ async function acquireGo(version: string, gotipRef: string, bootstrapGo: string)
         // Extract latest commit hash.
         commitHash = await executil.gitRevParse(gitDir, 'FETCH_HEAD');
       }
-      // Update cache for git dir.
-      gitDir = await tc.cacheDir(gitDir, 'gotip', 'master', '');
 
       // Avoid building multiple times by caching work tree.
-      let workTreeCache = tc.find('gotip', commitHash);
+      let workTreeCache = tc.find('gotip', `0.0.0-devel.${commitHash}`);
       if (workTreeCache) {
         workTree = workTreeCache;
       } else {
@@ -140,8 +144,9 @@ async function acquireGo(version: string, gotipRef: string, bootstrapGo: string)
           cmd = 'make.bat';
         }
         await exec.exec(cmd, undefined, { cwd, env });
-        // Update cache for work tree.
-        workTree = await tc.cacheDir(workTree, 'gotip', commitHash);
+
+        // Add cache for work tree.
+        workTree = await tc.cacheDir(workTree, 'gotip', `0.0.0-devel.${commitHash}`);
       }
       toolRoot = workTree;
   } else {
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -1,6 +1,12 @@
+import * as tempdir from './tempdir';
+import * as executil from './executil';
+
 // Load tempDirectory before it gets wiped by tool-cache
-let tempDirectory = process.env['RUNNER_TEMPDIRECTORY'] || '';
+const tempDirectory = tempdir.tempDir();
+
 import * as core from '@actions/core';
+import * as exec from '@actions/exec';
+import * as io from '@actions/io';
 import * as tc from '@actions/tool-cache';
 import * as os from 'os';
 import * as path from 'path';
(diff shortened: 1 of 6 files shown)


tie:
Ubuntu tests are passing, getting errors on Windows.

tie:
Aside from Windows errors, the code is working. Opening for review.

src/installer.ts
```diff
@@ -180,6 +224,27 @@ async function determineVersion(version: string): Promise<string> {
   return await getLatestVersion(version);
 }
 
+async function getLatestCommit(branch: string): Promise<string> {
+  let url: string = util.format('https://go.googlesource.com/go/+/refs/heads/%s?format=JSON', branch);
+  //
+  // We don’t use restm.RestClient because Gitiles returns )]}'
+  // garbage on the first line of (invalid) JSON response.
```
zikaeroh: The bad JSON is on purpose; I would recommend linking to the documentation for this: https://gerrit-review.googlesource.com/Documentation/rest-api.html#output

src/installer.ts
```diff
@@ -180,6 +224,27 @@ async function determineVersion(version: string): Promise<string> {
...
+  //
+  let http: httpm.HttpClient = new httpm.HttpClient('setup-go');
+  let resp: httpm.HttpClientResponse = await http.get(url);
+  let status: number | undefined = resp.message.statusCode;
+  if (status != 200) {
+    throw new Error(`HTTP request to ${url} failed with ${status} status code`);
+  }
+  let body: string = await resp.readBody();
+  let start: number = body.indexOf('{');
```
zikaeroh: The docs (linked above) say to just remove the first line. Might be simpler.

src/installer.ts
```diff
@@ -77,20 +80,57 @@ async function acquireGo(version: string): Promise<string> {
...
+      cwd: path.join(toolRoot, 'src'),
+      env: {
+        ...process.env,
+        // Embed in the executable, set it explicitly for reproducible builds.
+        'GOROOT_FINAL': '/go',
+      },
+    };
+    // TODO: is it OK to run make.bash on Windows?
+    // Anyway, I have no idea how to run make.bat using exec. Help would be appreciated.
+    let rc: number = await exec.exec('bash', ['make.bash'], options);
```
zikaeroh: Not okay to run `make.bash` on windows. Need to run `make.bat`. https://golang.org/doc/install/source#install

src/installer.ts
```diff
@@ -77,20 +80,57 @@ async function acquireGo(version: string): Promise<string> {
...
+    // Version embed in the executable. Required for build.
+    let versionFile = path.join(toolRoot, 'VERSION');
+    fs.writeFileSync(versionFile, `gotip-${version}`);
+
+    const options = {
+      cwd: path.join(toolRoot, 'src'),
+      env: {
+        ...process.env,
+        // Embed in the executable, set it explicitly for reproducible builds.
+        'GOROOT_FINAL': '/go',
```
zikaeroh: This path seems to be platform specific. Is this needed? I've never personally had to set this.

zikaeroh: Is `toolRoot` fine here? Since presumably that's a path for the host system.

src/installer.ts
```diff
@@ -77,20 +80,57 @@ async function acquireGo(version: string): Promise<string> {
...
+    version = normalizeVersion(version);
+  }
+
+  //
+  // Build gotip from source.
+  //
+  if (gotip) {
+    // Version embed in the executable. Required for build.
+    let versionFile = path.join(toolRoot, 'VERSION');
+    fs.writeFileSync(versionFile, `gotip-${version}`);
```
zikaeroh: This isn't quite the same as what building directly from a git clone would output. On stable, the equivalent to this file would be `go1.13.5`. On tip, it'd be `devel +<hash>`, where `<hash>` is the shortened git commit hash produced by `git rev-parse --short HEAD`. Unfortunately, I'm not sure you can get the shortened string without actually cloning as truncating a hash isn't the same logic, as `--short` means "give me at least this many unique characters" to deal with potential collisions.

zikaeroh:
I left some comments, but I'm still pretty iffy on the magic of using an API to fetch the commit, grabbing an archive, and creating a VERSION file, given how far it is from an actual build that people working with `tip` actually do.

The main use case of running `tip` in CI is to find issues ahead of time, be it on Go itself or in user code. For example, I found broken unsafe code in one of my dependencies because I ran race tests with `tip`, which enabled `checkptr` and revealed the issue.

Being too far removed from the version output expected by the Go devs, or the install process in the documentation I think may not be a good idea. I'd really consider another method where a plain `git clone --depth=1 ...` is used.

tie:
I’ve reconsidered my decision on using archives.

Plain `git` will downloads around 138M, while the archive size is 21M and 113M. The difference is negligible. Also, using `git` should fix issues with Windows.

We could cache `.git` separately and do `git pull --depth=1` for updates, which would be overall faster than downloading archive each time.

src/installer.ts
tie: Yeah, should remove the leading slash. We export `GOROOT` variable, so there is no reason to embed potentially different `GOROOT_FINAL` value on each build.

zikaeroh: Does setup-go always export GOROOT? This would explain the trouble I had getting tip setup. Typically, you don't need to set GOROOT at all, since the go toolchain will infer it. I'd be best to leave it unset, but that's another issue.

src/installer.ts
```diff
@@ -77,20 +80,57 @@ async function acquireGo(version: string): Promise<string> {
...
+
+    const options = {
+      cwd: path.join(toolRoot, 'src'),
+      env: {
+        ...process.env,
+        // Embed in the executable, set it explicitly for reproducible builds.
+        'GOROOT_FINAL': 'go',
```
zikaeroh: I guess I should be more clear on my question; what is this for? I don't need to set this to have a working build of Go. This is the first time I've seen this variable before.

tie: Default `GOROOT` value baked into Go binaries is either `GOROOT_FINAL` (if set) or build directory. I’d expect the latter to differ between builds on CI system. Go toolchain uses the `GOROOT` env var that we export explicitly, not the default one.

I’m keen on [reproducible builds](https://reproducible-builds.org) and Go team does an excellent job supporting these. Though we’d also have to compile with `CGO_ENABLED=0` for complete reproducibility, because otherwise external linker and Cgo embeds timestamps / paths in binaries. Note that disabling Cgo during compilation does not entirely disable Cgo, but instead just builds Go without using Cgo.

Again, there is no point in caching build outputs if they depend on the environment. And we want to cache gotip, as the build time is pretty slow compared to an average Go package (1m30s vs few seconds).

You could try running `GOROOT_FINAL=go CGO_ENABLED=0 ./make.bash` in different directories and the results of `shasum ../bin/go` should be the same.

I also recommend using `-trimpath` flag with Go ≥1.13 for building user packages (not the toolchain), i.e. set `GOFLAGS=-trimpath`.

```
-trimpath
	remove all file system paths from the resulting executable.
	Instead of absolute file system paths, the recorded file names
	will begin with either "go" (for the standard library),
	or a module path@version (when using modules),
	or a plain import path (when using GOPATH).
```

tie: What should be done is allowing overrides from `process.env`.

```js
env: {
  'GOROOT_FINAL': 'go',
  'CGO_ENABLED': '0',
  ...process.env,
}
```

zikaeroh: Personally, I wouldn't want any of these options set, as they would not normally be set in a typical Go compiler build. Opting into this behavior if you really want the Go compiler to have these things set is fine, but I'd really want to be as close to a typical build (or the downloaded builds as with the other versions in this action) as possible. I'm personally not too convinced that having the actual Go compiler be reproducible is a requirement for having the things it builds be reproducible; that seems like the actual use case of reproducible builds in CI.

Can you even set environment variables on pre-created actions? I thought you could only use `with`.

tie: >Can you even set environment variables on pre-created actions (without adding extra steps)? I thought you could only use with.

Haven’t tried yet, but why wouldn’t that work?

>Commands run in actions or steps can create, read, and modify environment variables.

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables

zikaeroh: Not really what I meant.

```yaml
- use: actions/setup-go@master
  with:
    go-version: tip
  env:
    CGO_ENABLED: '0'
```

etc

What would environment variables mean for an action like this which is supposed to be caching results? Especially if the toolcache library is just taking in `version` as its key?

I feel like if a user has a specific requirement for the Go toolchain other than its version, they're going to have to do something special themselves with the knowledge that a generalized action may not have all of the functionality they require by default (i.e. "if you want to change what the compiler is built with, no caching for you", making things work for the general user who has no knowledge of setup-go's internals).

tie: My point is that if you are using Cgo, you probably already have `CGO_ENABLED=1` env exported to clearly state that this package uses Cgo. It’s really, really confusing when large projects uses  Cgo dependencies and does not explicitly declare itself as using Cgo, because you can’t just `grep` for `import "C"`.

And GitHub Actions doc (a bit vaguely) states that `env` should work for pre-created actions.

tie: And adding `GOROOT_FINAL=go` just makes the build reproducible. That’s it.

zikaeroh: GOROOT_FINAL doesn't seem harmful, so I'm indifferent either way, I'm just saying that it's not a strict requirement for having an action that itself can produce a reproducible build.

> if you are using Cgo, you probably already have `CGO_ENABLED=1` env exported to clearly state that this package uses Cgo. 

I have not seen anyone do this, especially since the default is enabled. It's really not important for installing Go in this action, either way.

tie: >I have not seen anyone do this

That’s the problem.

I’ve clearly stated my reasoning above. Summing up the main points,

- if you are using Cgo, you probably already have `CGO_ENABLED=1`,
- there is no point in caching build outputs if they depend on the environment.

If you don’t want Cgo to be disabled by default (if that’s your objection), we can export `CGO_ENABLED=$($GOROOT_BOOTSTRAP/bin/go env CGO_ENABLED)` after the build. But I strongly oppose the implicit Cgo.

Also, there are no packages using `tip` version with [actions/setup-go@master](https://github.com/actions/setup-go). So we won’t break anything. And if you do indeed depend on some quirky Cgo feature, be it the external linker behavior, or plain `import "C"`, and want to test against `tip`, you are probably prepared for things to explode.

zikaeroh: I really can't agree. All disabling cgo is going to do is piss off every user of the action who (knowingly or not) depend on it, i.e. anyone with many database drivers like sqlite, embedded luajit, etc, and cause unnecessary "why doesn't this work" churn. I get that the average "tip" user in CI is going to know better, but I don't feel that needs to be related, especially if people want to switch from any other CI platform to Actions and discover "oh, this action is opinionated differently than every other CI so now I've got to do this other thing".

"I have not seen anyone do this" being a problem you're trying to solve, and "you probably already have GCO_ENABLED=1" are two competing thoughts. They cannot both be true.

> Also, there are no packages using tip version with actions/setup-go@master.

Well yeah, the functionality doesn't exist... Other than my previous comments which build it (and I run tip on my projects so I can report issues upstream).

Either way, your PR does not include any cgo changes, so whatever.

zikaeroh: To be clear, I'm just talking about the environment _after_ setup-go. If you want to build the Go compiler without cgo, then that's really not a major problem (other than it differing from running `make.bash` from a typical environment, and not necessarily being needed for said compiler to produce a consistent output; maybe it's necessary for proper caching between builds). I just don't want to see things getting quietly set for future steps in the pipeline.

tie: Sadly, looks like we won’t reach an agreement on this, as we both seem to have a strong opinion on the topic.

>I have not seen anyone do this

Sorry for being unclear in the previous comment, I meant that the problem is with packages not defining `CGO_ENABLED=1` in the environment. Personally, I have a negative experience working working with such projects, so my opinion is biased.

Also, I’ve seen people doing matrix builds for both `CGO_ENABLED=1` and `0`.

That said, maybe we can cache iff `CGO_ENABLED=0` and provide an opt-in for caching (non-reproducible) Cgo builds? Looks like good compromise to me.

zikaeroh: I don't understand what "packages not defining `CGO_ENABLED=1`" means; Go packages cannot specify the build environment they're intended to be built within. If it breaks, it breaks. Packages which intend to be built with either or need to guard themselves with build tags.

> Also, I’ve seen people doing matrix builds for both `CGO_ENABLED=1` and `0`.

This to me is evidence that setup-go should not set these variables for future steps (and only use internally as needed), as matrix builds set their environments at the start. An action overriding them defeats that.

Again, if you want to set `CGO_ENABLED=0` strictly for `make.bash`/`make.bat` as a part of building Go at `tip` and caching it between workers, then that's a different story, and I can understand wanting that to be consistent.

tie: Ah, maybe GitHub will add automatic updates for comments one day… Didn’t see the follow ups. Anyway, if disabling caches for `CGO_ENABLED=1` is not the problem, we can agree on this. Otherwise, it’s possible to export `CGO_ENABLED` from bootstrap toolchain for future steps and have a faster CI runs with Cgo.

tie: >Again, if you want to set CGO_ENABLED=0 strictly for make.bash/make.bat as a part of building Go at tip and caching it between workers, then that's a different story, and I can understand wanting that to be consistent.

Then we can finally settle on `CGO_ENABLED=$($GOROOT_BOOTSTRAP/bin/go env CGO_ENABLED)` for future steps and `CGO_ENABLED=0` for `make.bash`/`make.bat`. Yay!

zikaeroh:
```
CGO_ENABLED=$($GOROOT_BOOTSTRAP/bin/go env CGO_ENABLED)
```

I'm assuming you're talking about `::set-env` here; isn't this a no-op?

tie: It’s is, but only if using such matrix builds.

If `CGO_ENABLED` is not defined, then we default to whatever bootstrap toolchain sees fit for the host platform. Otherwise, as you’ve said earlier, we’d might change the default behavior of the toolchain.

tie:
Almost there on another branch, having a problem with lint/prettifier being not helpful at all (it just fails and tells to read the log _file_ for details).

tie:
Also closes #35 because I ended up refactoring tests and found this issue.

src/installer.ts
tie: >Does setup-go always export GOROOT?

It does. Not sure if it’s best left unset, especially in the CI environment.

src/installer.ts
tie: Fixed.

tie:
Ready for review again.

bryanmacfarlane:
Interested in these features potentially but master changed drastically with the v2-beta changes.  Do you want to take a stab at implementing on top of that code?

bryanmacfarlane:
Looking over this soon.  assigning to me

wyk9787:
Is there an estimated timeline when this will be implemented? This will be really helpful for us to transit into using Github Action.

tie:
Closing this PR since it’s out-of-date.

PR #40, opened 2020-02-09
v2 with proxy support

bryanmacfarlane:
- Proxy support
- Bug fixes with semver matching 
- stable option so you can specify not stable versions.
- Test coverage including mocking to cover actions code arcs (no internet needed)
- e2e workflow to validate downloading versions

cc


.github/workflows/versions.yml
```diff
@@ -0,0 +1,44 @@
+name: go-versions
+on:
+    push:
```
ericsciple: nit: looks like this has extra level of indent

README.md
```diff
@@ -1,14 +1,32 @@
...
+- uses: actions/checkout@v2
+- uses: actions/setup-go@v2-beta
+  with:
+    go-version: '^1.13.1' # The Go version to download (if necessary) and use.
+- run: go version
+```
+
 # Usage
 
 See [action.yml](action.yml)
```
ericsciple: nit: Consider adding the inputs to the readme, rather than pointing to action yaml. I have a script that can keep the two up to date (see [gendocs](https://github.com/actions/checkout/blob/master/package.json#L12) command in package.json; workflow validates `git diff` is empty)

src/installer.ts
```diff
@@ -1,223 +1,114 @@
...
-async function determineVersion(version: string): Promise<string> {
-  if (!version.endsWith('.x')) {
-    const versionPart = version.split('.');
-
-    if (versionPart[1] == null || versionPart[2] == null) {
-      return await getLatestVersion(version.concat('.x'));
-    } else {
-      return version;
+    debug(`check ${version} satisfies ${versionSpec}`);
+    if (semver.satisfies(version, versionSpec) && candidate.stable == stable) {
```
ericsciple: should this be `if (semver.satisfies(version, versionSpec) && (!stable || candidate.stable === stable))`

ericsciple: the description for stable is: `Whether to download only stable versions`

so i would expect if a customer specifies `stable: false`, it would be OK if they get a stable version

src/main.ts
```diff
@@ -0,0 +1,50 @@
...
+  try {
+    //
+    // versionSpec is optional.  If supplied, install / use from the tool cache
+    // If not supplied then problem matchers will still be setup.  Useful for self-hosted.
+    //
+    let versionSpec = core.getInput('go-version');
+
+    // stable will be true unless false is the exact input
+    // since getting unstable versions should be explicit
+    let stable = Boolean(core.getInput('stable') || 'true');
```
ericsciple: from a node repl:
```
> Boolean('false')
true
```

in checkout, i did this: `(core.getInput('stable') || 'true').toUpperCase() === 'TRUE'`

src/installer.ts
```diff
@@ -1,223 +1,114 @@
...
-    // handle beta and rc: 1.10beta1 => 1.10.0-beta1, 1.10rc1 => 1.10.0-rc1
-    if (versionPart[1].includes('beta') || versionPart[1].includes('rc')) {
-      versionPart[1] = versionPart[1]
-        .replace('beta', '.0-beta')
-        .replace('rc', '.0-rc');
-      return versionPart.join('.');
-    }
-  }
+  let goFile: IGoVersionFile | undefined;
+  for (let i = 0; i < candidates.length; i++) {
```
ericsciple: nit: `for (const candidate of candidates)` would simplifiy a little

src/installer.ts
bryanmacfarlane: Good catch.  I also need to add a test for that.  Yep, the expectation is that if they say false, then it gets the latest match on that semver spec regardless of stable.  In other words, true means we look at the stable attribute, false means we don't even consider the candidate.stable data.

src/main.ts
bryanmacfarlane: Will change.  And I'll add units / e2e around stable

__tests__/setup-go.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    });
+    logSpy.mockImplementation(line => {
+      // uncomment to debug
+      // process.stderr.write('log:' + line + '\n');
+    });
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+    jest.clearAllMocks();
```
ericsciple: @bryanmacfarlane From [the docs](https://jestjs.io/docs/en/mock-function-api.html#mockfnmockreset), it looks like reset does everything clear does and more

Do you need to call `jest.restoreAllMocks` during `afterAll`, which "restores the original (non-mocked) implementation."

PR #41, opened 2020-03-02
Add $GOBIN

### LINKED ISSUE #14: Add $GOPATH/bin to $PATH

fsouza:
- Set GOBIN to $(go env GOPATH)/bin
- Add GOBIN to the PATH

Should make the setup of tools like golangci-lint or golint work with a
simple `go get`.

Using $GOBIN instead of $GOPATH/bin because the goal is to have GOPATH
not being directly referenced. Also, in the future, GOBIN will have a
default value too, so we would not need to manually set it, just add it
to the path (see discussion in golang/go#23439).

Closes #14.

### DIFF
--- a/.github/workflows/versions.yml
+++ b/.github/workflows/versions.yml
@@ -19,26 +19,26 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
-    - name: setup-go ^1.13.6
+    - name: setup-go ^1.14.0
       uses: ./
       with:
-        go-version: ^1.13.6
+        go-version: ^1.14.0
 
-    - name: validate version 
-      run: go version | grep "go1.13."
+    - name: validate version
+      run: go version | grep "go1.14."
 
     - name: setup-go 1.13
       uses: ./
       with:
         go-version: 1.13
 
-    - name: validate version 
-      run: go version | grep "go1.13."      
+    - name: validate version
+      run: go version | grep "go1.13."
 
     - name: setup-go 1.12.9
       uses: ./
       with:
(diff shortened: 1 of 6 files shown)


fsouza:
To see how this can be used: https://github.com/fsouza/vod-module-sprite/pull/15/files

(Also notice the previous workaround in the "setup GOBIN" step).

Commit 25c870be4dd26e41bb94c88b9912380fb18459e8 caused the tests to fail because what seemed to be an incorrec test. "1.14.0" satisfies "^1.13.6" (and it looks like semver doesn't support the ~ operator yet, which is likely what the test is designed to test?)

I adjusted the test to make it green (using ^1.14.0 instead of ^1.13.6), but if we keep that test as is, it's going to fail every time there's a new "minor" release of Go.

fsouza:
@bryanmacfarlane friendly ping? 😁

mingrammer:
I'm waiting for this feature :)

fsouza:
Implemented in #47.

bryanmacfarlane:
Thank you @fsouza for raising the issue.

PR #43, opened 2020-03-07
Update setup-go per go1.14 release

chenrui333:

PR #44, opened 2020-03-08
Update jest-ts and update audit level

chenrui333:
Split from #43 

cc @bryanmacfarlane


bryanmacfarlane:
Thanks for the PR but another PR and change fixed this.

PR #47, opened 2020-03-26
add bin to path

bryanmacfarlane:
fix issue with cache dir (had spec instead of just version)
add bin to path 
fix audit issues
update tool-cache lib

Details:
This bin dir is an output dir when you build + install something in go.  So users want to call go build, go install and then run what they built.  This adds that output dir to the path.

hosted images and some folks install go but if they never did a build + install on a clean machine (hosted) it won’t exist.  That’s why I create it if (1) go is installed and (2) it doesn’t exist

To avoid fragility I exec go env GOPATH and let it tell me where it should be

Crux of the change: https://github.com/actions/setup-go/pull/47/files#diff-8cfead41d88ad47d44509a8ab0a109ad

Screen Shot 2020-03-26 at 1 49 59 PM

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -63,7 +63,7 @@ export async function run() {
 async function addBinToPath(): Promise<boolean> {
   let added = false;
   let g = await io.which('go');
-  core.debug(`which go :${g};`);
+  core.debug(`which go :${g}:`);
   if (!g) {
     core.debug('go not in the path');
     return added;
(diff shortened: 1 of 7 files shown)


src/main.ts
```diff
@@ -48,3 +54,34 @@ export async function run() {
...
+  let g = await io.which('go');
+  core.debug(`which go :${g}:`);
+  if (!g) {
+    core.debug('go not in the path');
+    return added;
+  }
+
+  let buf = cp.execSync('go env GOPATH');
+  if (buf) {
+    let gp = buf.toString().trim();
```
ericsciple: any concern about the output? for example should we check empty string

Issue #46, opened 2020-03-23
Matchers: Error annotations catch things that are not errors and not even in Go code


JasonGross:
Consider the (totally valid) makefile:
```makefile
all: foo
	echo Done

include Foo.mk

Foo.mk:
        printf 'foo:\n\techo foo\n' > $@
```
On the first run of `make`, this generates
```
Makefile:4: Foo.mk: No such file or directory
printf 'foo:\n\techo foo\n' > Foo.mk
echo foo
foo
echo Done
Done
```
There isn't really a good way to suppress this warning (see [this StackOverflow question I asked](https://stackoverflow.com/questions/43920369/suppress-make-warning-from-not-yet-built-include-without-suppressing-error)) without massively complicating the Makefile.  I don't mind seeing this warning on the first `make`.  However, my github actions go job decides that this is an error (it's not; it's non-fatal), and annotates *every single pull request* on my repository with an error on the relevant line on the Makefile.  Please fix the annotation to not do this.

bryanmacfarlane:
I think this action does the right thing.  Perhaps the issue is service side which turns output into warning and error annotations.  @joshmgross to verify whether there's anything we can do action side.

JasonGross:
No, I'm pretty sure this is an issue with your matcher being overzealous:
https://github.com/actions/setup-go/blob/1c06f0e82edaef79c0f495bae346cd3f650f4ef4/matchers.json#L7

This claims that every single output line containing two `:`s separated by a number an followed by a space is an error in the file that comes before the first colon.

bryanmacfarlane:
Ahhh.  OK.  Thanks @JasonGross 

@thboop has another issue on the matcher so maybe he can look at both together.

bryanmacfarlane:
@JasonGross if it's causing problems there's also a way to disable the problem matcher.  

https://github.com/actions/toolkit/blob/master/docs/commands.md#problem-matchers

We added that feature since the scope of a problem matcher is a job and scripts / steps downstream could get false hits.

bryanmacfarlane:
I think the solution is to have a much tighter regex for compile errors and then if there's things it misses, problem matchers supports having multiple matchers.

I'll have a PR soon for the v2-beta lineage.

JasonGross:
> @JasonGross if it's causing problems there's also a way to disable the problem matcher.
> 
> https://github.com/actions/toolkit/blob/master/docs/commands.md#problem-matchers
> 
[...]

PR #48, opened 2020-03-27
More focused problem matcher regex

bryanmacfarlane:
see: https://github.com/actions/setup-go/issues/46#issuecomment-604937131

The main change in this PR is the regex pattern in `matcher.json`.  However, I added some tests to validate it catches cases.

Instead of creating one general and broad regex to catch everything, our problem matchers should be more focused and if there's cases we miss, the platform supports adding many problem matchers.

setup-go is also about to release `v2` with the current tag being `v2-beta` so this is the perfect time to try something more restrictive.  If we find over time there's other error cases, we will add additional patterns.


__tests__/setup-go.test.ts
```diff
@@ -9,6 +9,9 @@ import * as main from '../src/main';
 import * as im from '../src/installer';
 
 let goJsonData = require('./data/golang-dl.json');
+let matchers = require('../matchers.json');
+let matcherPattern = matchers.problemMatcher[0].pattern[0];
+let matcherRegExp = new RegExp(matcherPattern.regexp);
```
ericsciple: cool

matchers.json
```diff
@@ -4,11 +4,11 @@
             "owner": "go",
             "pattern": [
                 {
-                    "regexp": "^([^:]*: )?((.:)?[^:]*):(\\d+)(:(\\d+))?: (.*)$",
-                    "file": 2,
-                    "line": 4,
-                    "column": 6,
-                    "message": 7
+                    "regexp": "((?:\\.{0,2}\\/|\\.{0,2}\\\\)(?:.+\\.go)):(?:(\\d+):(\\d+):)? (.*)",
```
ericsciple: nit: It looks like you might be able to eliminate some groups. I think this is equivalent:

```
"(\\.{0,2}[\\/\\\\].+\\.go):(?:(\\d+):(\\d+):)? (.*)",
```

- moved `\\.{0,2}` in front of the group
- then switched `(?:\\/|\\\\)` to a char set `[\\/\\\\]` - arguably simpler? idk
- Removed group around the rest of the file path `(?:.+\\.go)` =\> `.+\\.go`

__tests__/setup-go.test.ts
```diff
@@ -322,6 +325,100 @@ describe('setup-go', () => {
...
+    let line = '..\\main.go:13:2: undefined: fmt.Printl';
+    let annotation = testMatch(line);
+    expect(annotation).toBeDefined();
+    expect(annotation.line).toBe(13);
+    expect(annotation.column).toBe(2);
+    expect(annotation.file).toBe('..\\main.go');
+    expect(annotation.message).toBe('undefined: fmt.Printl');
+  });
+
+  it('only matches leading dots on unix path', async () => {
```
ericsciple: The description doesnt align with the behavior

Should the description be updated, or should the behavior be fixed?

matchers.json
bryanmacfarlane: Yeah, I think that is better.  I'll change

__tests__/setup-go.test.ts
ericsciple: I got confused why a match was returned at all.

The regexp doesn't have `^$` around the pattern. Should it?

bryanmacfarlane: I think it matches??  In this line:
```
x./assert.go:5:2
```
it's validating that it only matches the leading dot(s) and not the x.  Not sure this output would ever happen but my initial regex had a bug where I wasn't escaping the dot so it was match 0-2 of any char in from of the slash which is not correct.

Do you have a better suggestion for the title?

matchers.json
bryanmacfarlane: I changed it to this.

__tests__/setup-go.test.ts
bryanmacfarlane: fixed.

PR #58, opened 2020-06-23
Use GitHub releases to download Go versions.

dmitry-shibanov:
This pull-request improves setup-go action to add ability to download specific version of Go from [actions/go-versions](https://github.com/actions/go-versions) releases on flight if it is not available by default.
## Details:
We added intermediate step to download go from releases and current setup has the following order:
```
1. Try to find version locally
2. If it's not found locally, try to find in release manifest
3. If version is found in manifest download from releases
4. if version is not found in releases try to install from dist.
```
Additional changes:

1. Move archive extract logic to the separate function to improve readability
2. Move installation logic to common function
3. Extend test matrix in versions.yml

**Note:** unit tests are in progress.

### DIFF
--- a/__tests__/setup-go.test.ts
+++ b/__tests__/setup-go.test.ts
@@ -3,7 +3,7 @@ import * as io from '@actions/io';
 import * as tc from '@actions/tool-cache';
 import fs from 'fs';
 import cp from 'child_process';
-import osm = require('os');
+import osm from 'os';
 import path from 'path';
 import * as main from '../src/main';
 import * as im from '../src/installer';
--- a/__tests__/setup-go.test.ts
+++ b/__tests__/setup-go.test.ts
@@ -67,7 +67,7 @@ describe('setup-go', () => {
 
     // writes
     cnSpy = jest.spyOn(process.stdout, 'write');
-    logSpy = jest.spyOn(console, 'log');
+    logSpy = jest.spyOn(core, 'info');
     dbgSpy = jest.spyOn(core, 'debug');
     getSpy.mockImplementation(() => <im.IGoVersion[] | null>goJsonData);
     cnSpy.mockImplementation(line => {
--- a/__tests__/setup-go.test.ts
+++ b/__tests__/setup-go.test.ts
@@ -10,6 +10,7 @@ import * as im from '../src/installer';
 
 let goJsonData = require('./data/golang-dl.json');
 let matchers = require('../matchers.json');
+let goTestManifest = require('./data/versions-manifest.json');
 let matcherPattern = matchers.problemMatcher[0].pattern[0];
 let matcherRegExp = new RegExp(matcherPattern.regexp);
 
@@ -32,6 +33,8 @@ describe('setup-go', () => {
   let existsSpy: jest.SpyInstance;
   let mkdirpSpy: jest.SpyInstance;
   let execSpy: jest.SpyInstance;
+  let getManifestSpy: jest.SpyInstance;
+  let getInfoManifestSpy: jest.SpyInstance;
 
   beforeEach(() => {
     // @actions/core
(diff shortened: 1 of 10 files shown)


bryanmacfarlane:
👀

__tests__/setup-go.test.ts
```diff
@@ -52,18 +54,22 @@ describe('setup-go', () => {
...
 
+    // gets
+    getManifestSpy.mockImplementation(() => <tc.IToolRelease[]>goTestManifest);
+
     // writes
     cnSpy = jest.spyOn(process.stdout, 'write');
     logSpy = jest.spyOn(console, 'log');
-    dbgSpy = jest.spyOn(main, '_debug');
-    getSpy.mockImplementation(() => <im.IGoVersion[]>goJsonData);
+    dbgSpy = jest.spyOn(core, 'debug');
```
bryanmacfarlane, line 71: Did this actually mock out the debug output or were you seeing it in the output?  I was having an issue with that.  If so, then great- this is cleaner

src/installer.ts
```diff
@@ -1,46 +1,12 @@
 import * as tc from '@actions/tool-cache';
+import * as core from '@actions/core';
 import * as path from 'path';
 import * as semver from 'semver';
 import * as httpm from '@actions/http-client';
 import * as sys from './system';
-import {debug} from '@actions/core';
+import os from 'os';
```
bryanmacfarlane, line 7: Did you get the os to mock successfully?  In setup-node I had to do require('os') because that's the way the lib did it and it would only mock if it was loaded from the node loader in the same way.  If so, curious what the difference was.

bryanmacfarlane:
I had a couple questions / comments but I'll approve.  We can follow up on those async.

src/installer.ts
```diff
@@ -55,6 +21,153 @@ export interface IGoVersion {
...
+
+async function installGoVersion(
+  info: IGoVersionInfo,
+  auth: string | undefined
+): Promise<string> {
+  console.log(`Acquiring ${info.resolvedVersion} from ${info.downloadUrl}`);
+  const downloadPath = await tc.downloadTool(info.downloadUrl, undefined, auth);
+
+  console.log('Extracting Go...');
+  let extPath = await extractGoArchive(downloadPath);
```
konradpabjan, line 108: I would improve the logging here a little bit.

I would add something like `core.info('Successfully extracted go to ${extPath }')` after the extraction is complete.

Something similar after `tc.cacheDir` would also be nice.

Once thing thing that I noticed in `setup-python` is that it's better to have more output & logging vs less. There have been some issues reported and it's much easier to diagnose them when there is more output.

dmitry-shibanov: Thank you, we changed log output and added extra output after `tc.cacheDir`.

src/installer.ts
```diff
@@ -55,6 +21,153 @@ export interface IGoVersion {
...
+) {
+  let osPlat: string = os.platform();
+  let osArch: string = os.arch();
+
+  // check cache
+  let toolPath: string;
+  toolPath = tc.find('go', versionSpec);
+  // If not found in cache, download
+  if (toolPath) {
+    console.log(`Found in cache @ ${toolPath}`);
```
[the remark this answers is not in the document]
dmitry-shibanov: Done.

src/main.ts
```diff
@@ -23,28 +23,18 @@ export async function run() {
...
-        let added = addBinToPath();
-        core.debug(`add bin ${added}`);
-      } else {
-        throw new Error(
-          `Could not find a version that satisfied version spec: ${versionSpec}`
-        );
-      }
+      let added = addBinToPath();
+      core.debug(`add bin ${added}`);
+      console.log('Done');
```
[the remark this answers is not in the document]
dmitry-shibanov: Done.

src/main.ts
```diff
@@ -23,28 +23,18 @@ export async function run() {
...
-      }
+      const installDir = await installer.getGo(versionSpec, stable, auth);
 
-      if (installDir) {
-        core.exportVariable('GOROOT', installDir);
-        core.addPath(path.join(installDir, 'bin'));
-        console.log('Added go to the path');
+      core.exportVariable('GOROOT', installDir);
+      core.addPath(path.join(installDir, 'bin'));
+      console.log('Added go to the path');
```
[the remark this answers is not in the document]
dmitry-shibanov: Thank you, We replaced console.log to core.info

src/installer.ts
dmitry-shibanov: Yes, it mocks successfully.

__tests__/setup-go.test.ts
dmitry-shibanov: We want to clarify one point, as I understand that we should not receive output while these lines are commented out?
```
    dbgSpy.mockImplementation(msg => {
      // uncomment to see debug output
      // process.stderr.write(msg + '\n');
    });
```
If we uncomment them, should we get a conclusion or not? On normal startup, we do not get debug output.

PR #75, opened 2020-09-14
Add `version-resolver` input

### LINKED ISSUE #73: Version in tool cache is often behind go-versions manifest's latest
## Goal: Set minimum required Go version but pick the latest available
For the use case of specifying a lower bound Go version for a project but defaulting to the latest release I'm using this action config fragment:

```
      - uses: actions/setup-go@v2
        with:
          go-version: '>=1.5 <2'
```

### Check tool-cache
setup-go first checks if a matching version Go is in the tool cache. https://github.com/actions/setup-go/blob/main/src/installer.ts#L41

With `ACTIONS_STEP_DEBUG=true`, the relevant part from a build run:
```
Setup go stable version spec >=1.5 <2
##[debug]isExplicit: 
##[debug]explicit? false
##[debug]isExplicit: 1.11.13
##[debug]explicit? true
##[debug]isExplicit: 1.12.17
##[debug]explicit? true
##[debug]isExplicit: 1.13.15
##[debug]explicit? true
##[debug]isExplicit: 1.14.7
##[debug]explicit? true
##[debug]isExplicit: 1.15.0
##[debug]explicit? true
##[debug]evaluating 5 versions
##[debug]matched: 1.15.0
##[debug]checking cache: /opt/hostedtoolcache/go/1.15.0/x64
##[debug]Found tool in cache go 1.15.0 x64
Found in cache @ /opt/hostedtoolcache/go/1.15.0/x64
```
_(Although the debug looks like it's iterating from oldest version in cache, it actually isn't the behavior. These are `_isExplicitVersion(...)` calls from `findAllVersions()` in `@actions/toolkit/.../tool-cache.ts`
https://github.com/actions/toolkit/blob/master/packages/tool-cache/src/tool-cache.ts#L514)_

### GitHub local Go Versions Manifest
Only if there isn't a matching version in the cache, it proceeds to checking `@actions/go-versions:versions-manifest.json`.  A file that seems to be periodically generated, approved by humans in a [PR](https://github.com/actions/go-versions/pull/14), and gets committed again by a bot.

If we check versions-manifest.json, [it has Go 1.15.1 in it](https://github.com/actions/go-versions/blob/817f326783f197b277c57e9e9c58277376b8067e/versions-manifest.json#L3). On the other hand the cache only has—semver cleaned up—1.15.0.

### Can we prime the cache?
To debug this, I tried priming the cache by hardcoding the version and seeing if [`installGoVersion()`](https://github.com/actions/setup-go/blob/8a3a76c2171de8c3be20bec507b6d829ccae48ba/src/installer.ts#L100) from `installer.ts` would actually put this into the cache.

It says so:
```
##[debug]Caching tool go 1.15.1 x64
##[debug]source dir: /home/runner/work/_temp/9b74f02e-5739-4ef8-8416-f3c2b222710b
##[debug]destination /opt/hostedtoolcache/go/1.15.1/x64
##[debug]finished caching tool
```
...but when I revert back to `>=1.5 <2` version spec and run the action again, it continues to match `1.5.0` from the tool-cache.

### How does /opt/hostedtoolcache get populated?
Seems like only PyPy, Python, Ruby, boost, go, and node are the tools in this directory.
Makes me think this is centrally curated, and not a globally shared cache.

### What cache gets updated by `installGoVersion()`?
When it calls [`tc.cacheDir`](https://github.com/actions/setup-go/blob/8a3a76c2171de8c3be20bec507b6d829ccae48ba/src/installer.ts#L115) on the extracted Go install, what is the effective cache domain for this? Is it global, per project, per user, per run?

### Is checking the cache with a version spec intentional?
If we don't have an explicit version, setup-go checks the cache with the supplied semver version spec. Is this intentional? Because of this, there's no way to provide a version spec but hope to expect to get the latest version in `go-versions` manifest file, or directly from Google.

bdd:
Tool cache can be stale in GitHub and self-hosted runners. If supplied
`go-version` is a semver range--instead of an explicit version--it will
be tried through the version inventory in cache, first. Even though
the GitHub local copy or origin distribution may have newer versions
matching the passed range, as long as the cache can satisfy the range,
the latest version from cache will be used.

Introduce an optional input, `version-resolver`. If defined, it tries to
resolve the version spec against either GitHub local copy or Go's
canonical source of version manifest.
If the value of `version-resolver` is:
  - "manifest": the manifest file under @actions/go-version get used.
  - "dist": the manifest at https://golang.org/dl/?mode=json&include=all
  gets used.  It can take values

Example:
--------
- Latest Go Version: 1.15.2
- tool-cache has go1.15 (1.15.0), latest.

build.yml:
```
[...]
  - uses: actions/setup-go@v2
    with:
      go-version: '>=1.5 <2'
[...]
```

Although the intention of the user might be to get latest
version between "oldest Go 1.15.0, newest--not inclusive--Go 2.0", the
cache will match this range with Go 1.15.0.

build.yml (v2):
```
[...]
  - uses: actions/setup-go@v2
    with:
      go-version: '>=1.5 <2'
      version-resolver: 'manifest'
[...]
```

With supplied version resolver, the semver range will be checked against
the local GitHub version manifest from @actions/go-versions[1], and
match 1.15.2. Cache will be queried with the resolved version, instead
of the range. When the cache gets updated globally, next runs will use
the tool from the cache, instead of downloading locally or in case of
resolver 'dist', directly from Google.

[1]: https://raw.githubusercontent.com/actions/go-versions/main/versions-manifest.json

Closes #73

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,9 +1,11 @@
 name: 'Setup Go environment'
 description: 'Setup a Go environment and add it to the PATH'
 author: 'GitHub'
-inputs: 
+inputs:
   go-version:
     description: 'The Go version to download (if necessary) and use. Supports semver spec and ranges.'
+  version-resolver:
+    description: 'Version inventory for resolving semver ranges before checking tool cache. Use "manifest" for @actions/go-versions, or "dist" for golang.org/dl'
   stable:
     description: 'Whether to download only stable versions'
     default: 'true'
--- a/dist/index.js
+++ b/dist/index.js
@@ -1502,7 +1502,21 @@ function run() {
             if (versionSpec) {
                 let token = core.getInput('token');
                 let auth = !token || isGhes() ? undefined : `token ${token}`;
-                const installDir = yield installer.getGo(versionSpec, stable, auth);
+                // Tool cache can be stale in GitHub and self-hosted runners. If supplied
+                // `go-version` is a semver range--instead of an explicit version--it will
+                // be tried through the version inventory in cache, first. Even though
+                // the GitHub local copy or origin distribution may have newer versions
+                // matching the passed range, as long as the cache can satisfy the range,
+                // the latest version from cache will be used.
+                //
+                // Allow passing a prefered version inventory for resolving versionSpec.
+                // - "manifest": GitHub hosted (@actions/go-versions)
+                // - "dist": Origin (https://golang.org/dl/?mode=json&include=all)
+                let resolver = core.getInput('version-resolver');
+                if (resolver && resolver != 'manifest' && resolver != 'dist') {
+                    throw new Error(`Unknown version-resolver: '${resolver}'`);
+                }
+                const installDir = yield installer.getGo(versionSpec, resolver, stable, auth);
                 core.exportVariable('GOROOT', installDir);
                 core.addPath(path_1.default.join(installDir, 'bin'));
                 core.info('Added go to the path');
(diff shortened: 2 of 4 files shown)


fAtaboada:
art point set sound point second a ax ap in roles the second time a gpa a second t the second fourth and the second in a number possible the second thrid is open to see the cot the cancer and the evdince of knowing a secondf set name and number tences to iopen A and to open A B C  D and then all points are open and send open E to D so  D can open and closed all poiints in a O N act point at 10 points 
PosterMaker2021_07_02_10_28_43

fAtaboada:
PosterMaker2021_07_02_10_30_17

fAtaboada:
PosterMaker2021_07_02_10_25_40

fAtaboada:
PosterMaker2021_07_02_10_44_55

fAtaboada:
PosterMaker2021_07_02_10_41_58

fAtaboada:
PosterMaker2021_07_02_10_28_23

fAtaboada:
PosterMaker2021_07_02_10_27_09

fAtaboada:
PosterMaker2021_07_02_10_36_47

fAtaboada:
PosterMaker2021_07_02_10_46_55

fAtaboada:
PosterMaker2021_07_02_10_48_08

fAtaboada:
PosterMaker2021_07_02_10_28_43

dmitry-shibanov:
Hello @bdd. The [check-latest](https://github.com/actions/setup-go/pull/186) functionality was added in this [release](https://github.com/actions/setup-go/releases/tag/v2.2.0). For now I'm going to close the pull request.

PR #89, opened 2020-12-04
Add GOBIN to PATH regardless of whether a version spec is specified

### LINKED ISSUE #49: Feature request: add GOBIN to PATH even if a version isn't installed
Request: setup-go should always add GOBIN to PATH, even if it uses the default Go installed in the environment. This would mirror how it always sets up Go matchers.

Per thread here: https://github.com/actions/setup-go/issues/39#issuecomment-607420683

Thanks!

leighmcculloch:
### What

Add GOBIN to PATH regardless of whether a version spec is specified.

### Why

The setup-go action adds the GOBIN to the PATH, but only if a version spec for Go is provided in the workflow configuration. When a version spec is provided the action ensures that version is installed, and then also that the GOBIN path is included in the PATH. The code that includes the GOBIN in the PATH has no dependency on the logic to install a specific version of Go and I can't see a reason why it can't run regardless. When it doesn't run, binaries that are installed in GOBIN by `go get` are not found by following steps in a workflow.

Fix #49

### DIFF
--- a/__tests__/setup-go.test.ts
+++ b/__tests__/setup-go.test.ts
@@ -441,7 +441,7 @@ describe('setup-go', () => {
       return '/Users/testuser/go';
     });
 
-    mkdirpSpy.mockImplementation(async () => { });
+    mkdirpSpy.mockImplementation(async () => {});
     existsSpy.mockImplementation(path => {
       return false;
     });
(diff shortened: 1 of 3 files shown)


leighmcculloch:
I couldn't see where the tests were when I first made the change, but I see now they are in the __tests_ directory. I'll circle back on this soon to add a test that passes with the change and fails without it.

leighmcculloch:
I've added tests that pass with the change and fail without it.

leighmcculloch:
@bryanmacfarlane 👋 Looking for 👀 on this PR. Could this be accepted?

jayvdb:
ping @joshmgross

jayvdb:
ping @bryanmacfarlane

jayvdb:
ping @dmitry-shibanov ; any chance you could rescue this?

dmitry-shibanov:
Hello @jayvdb. Sorry for the late response. I'm not expert in the GO. Can it bring breaking changes ?

jayvdb:
No it cant bring breaking changes.

dmitry-shibanov:
Thank you for your response @jayvdb. I think in this case we can recreate this pull request and take a look at it.

walles:
I'm running into this, any update?

PR #98, opened 2021-01-16
Problem matchers: Don't require relative paths to start with ./ or ../

dominikh:
Errors reported by Go, even when containing relative paths, do not
always begin with ./ or ../ – for example:

	$ go build ./...
	# honnef.co/go/tools/lintcmd
	lintcmd/format.go:28:8: syntax error: cannot use path := filepath.Clean(pos.Filename) as value

We could require either ./, ../ or at least one path separator and
still match Go's output. However, commonly used linters (such as
Staticcheck and golint) never use ./ for relative paths. Their output
stopped being matched when we moved from v1 to v2. I believe that
being able to match the output of linters is worth relaxing the
pattern for.

This change slightly relaxes the stricter pattern that was introduced
as part of v2 to address #46. However, the pattern is still stricter
than it was in v1 and as strict as it can be for most users.


dominikh:
Friendly ping

marten-seemann:
Not sure who's maintaining this repo. @joshmgross, @thboop or @bryanmacfarlane maybe?
Would it be possible to the this pull request merged some time?

dominikh:
@marten-seemann In the meantime I can recommend https://github.com/WillAbides/setup-go-faster

bryanmacfarlane:
Any idea on how the regex compares to the one VS Code uses?  I believe that's where the original one was pulled.

bryanmacfarlane:
related setup-node PR which compared to and pulled from vs code: https://github.com/actions/setup-node/pull/125

mvdan:
@bryanmacfarlane I don't think the author or anyone else in this thread is involved with VS Code. I hope it's clear that, from the original post here, the current regular expression isn't right. I personally don't mind what the final fix ends up being, but we do need a fix sooner than later :)

dominikh:
Exasperated ping.

q464pp:
可以拉取合併嗎？

bryanmacfarlane:
@maxim-lobanov can ya'll take a look at this PR?

maxim-lobanov (approved):
@dmitry-shibanov @MaksimZhukov could you please proceed with accepting these changes if they look good to you.

Sergey-Murtazin (approved):

vsafonkin (approved):

dmitry-shibanov (approved):

PR #114, opened 2021-06-14
Revert "Swap to Environment Files"

MaxFox196:
Reverts actions/setup-go#76


joshmgross:
👋  Hey @MaxFox196, environment files are the intended way going forward to communicate with the runner, so we don't want to revert this change.

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

PR #141, opened 2021-08-19
Fix extractor selection on windows

paulcacheux:
On Windows the extractor uses `tar` instead of `zip` because the code is using `os.arch()` (which returns the arch like x64 or arm64) instead of `os.platform()`.
This PR fixes this problem.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -5084,9 +5084,9 @@ function installGoVersion(info, auth) {
 }
 function extractGoArchive(archivePath) {
     return __awaiter(this, void 0, void 0, function* () {
-        const arch = os_1.default.arch();
+        const platform = os_1.default.platform();
         let extPath;
-        if (arch === 'win32') {
+        if (platform === 'win32') {
             extPath = yield tc.extractZip(archivePath);
         }
         else {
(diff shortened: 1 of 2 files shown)


Xxxxx1111:
#142

paulcacheux:
Here is a demo repo to demonstrate the problem: https://github.com/paulcacheux/poc-setup-go-demo

paulcacheux:
Friendly ping, maybe @joshmgross 
Is this repo still maintained ? It seems to be infected with bots..

joshmgross:
👋  Hey @paulcacheux, thanks for the ping. I'll find a team to review and release this bug fix.

joshmgross (approved):

paulcacheux:
The job that is failing seems to be unrelated to the changes. To fix it I think another PR updating the node dependencies would be needed in my opinion. Happy to help if needed

mmorel-35:
Hi @paulcacheux, I agree with you on 
>The job that is failing seems to be unrelated to the changes. 

I created #144 where I fixed those issues

joshmgross:
This fix should now be available in https://github.com/actions/setup-go/releases/tag/v2.1.4, and `v2` soon.

timhaines:
Great to see this. 💯   Would have saved me 20 minutes today if it was on `v2` already.

PR #146, opened 2021-08-24
Add release action version workflow

joshmgross:
This PR adds a workflow to automatically update tags when releasing new versions of this action, using the https://github.com/actions/publish-action action.

### DIFF
--- a/.github/workflows/release-new-action-version.yml
+++ b/.github/workflows/release-new-action-version.yml
@@ -1,0 +1,27 @@
+name: Release new action version
+on:
+  release:
+    types: [released]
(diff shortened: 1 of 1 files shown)


.github/workflows/release-new-action-version.md
```diff
@@ -0,0 +1,27 @@
+name: Release new action version
```
mmorel-35, line 1: Hi ! Shouldn't it be `release-new-action-version.yml` instead of `release-new-action-version.md`

joshmgross: Fixed! Thanks!

PR #184, opened 2022-01-20
Add check latest functionality

Sergey-Murtazin:
**Description:**
Add `check-latest` flag which forces the action to check if the cached version is the latest one.

**Related issue:**
[Version in tool cache is often behind go-versions manifest's latest #73](https://github.com/actions/setup-go/issues/73)

**Check list:**


Sergey-Murtazin:
Need to fix new three tests

Issue #169, opened 2021-11-21
AssertionError [ERR_ASSERTION]: a path argument must be provided


Hunter-Thompson:
**Description:**
While running setup-go, it fails to set the GOPATH

**Action version:**
v2

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**
v1.7

**Repro steps:**  
https://github.com/Hunter-Thompson/gojen/runs/4276146794?check_suite_focus=true

**Expected behavior:**
My go build command to run.

**Actual behavior:**

Error log:

```
 Setup go stable version spec 1.7
Attempting to download 1.7...
matching 1.7...
Not found in manifest.  Falling back to download directly from Go
Install from dist
Acquiring go1.7.6 from https://storage.googleapis.com/golang/go1.7.6.linux-amd64.tar.gz
Extracting Go...
/usr/bin/tar xz --warning=no-unknown-keyword -C /home/runner/work/_temp/a2859f92-5cf0-4750-8a1e-85b46144016f -f /home/runner/work/_temp/b124ab7e-e40a-421f-99ce-13f9035024bf
Successfully extracted go to /home/runner/work/_temp/a2859f92-5cf0-4750-8a1e-85b46144016f
Adding to the cache ...
Successfully cached go to /opt/hostedtoolcache/go/1.7.6/x64
Added go to the path
(node:1563) UnhandledPromiseRejectionWarning: AssertionError [ERR_ASSERTION]: a path argument must be provided
    at Object.<anonymous> (/home/runner/work/_actions/actions/setup-go/v2/dist/index.js:4752:18)
    at Generator.next (<anonymous>)
    at /home/runner/work/_actions/actions/setup-go/v2/dist/index.js:4692:71
    at new Promise (<anonymous>)
    at module.exports.672.__awaiter (/home/runner/work/_actions/actions/setup-go/v2/dist/index.js:4688:12)
    at Object.mkdirP (/home/runner/work/_actions/actions/setup-go/v2/dist/index.js:4751:12)
    at Object.<anonymous> (/home/runner/work/_actions/actions/setup-go/v2/dist/index.js:204:22)
    at Generator.next (<anonymous>)
    at /home/runner/work/_actions/actions/setup-go/v2/dist/index.js:57:71
    at new Promise (<anonymous>)
[...]

PR #211, opened 2022-03-28
await io.mkdirP

Frassle:
**Description:**
io.mkdirP returns a promise that should be awaited. Without this the promise runs in the background and `addBinToPath` returns OK but then the promise errors later on.

**Related issue:**
https://github.com/actions/setup-go/issues/169

**Check list:**

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -73,13 +73,13 @@ export async function addBinToPath(): Promise<boolean> {
     if (!fs.existsSync(gp)) {
       // some of the hosted images have go install but not profile dir
       core.debug(`creating ${gp}`);
-      io.mkdirP(gp);
+      await io.mkdirP(gp);
     }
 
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -73,13 +73,13 @@ export async function addBinToPath(): Promise<boolean> {
     if (!fs.existsSync(gp)) {
       // some of the hosted images have go install but not profile dir
       core.debug(`creating ${gp}`);
-      io.mkdirP(gp);
+      await io.mkdirP(gp);
```
brcrista, line 76: Thanks @Frassle!  Could you update the `dist/` file as well so CI passes?  You can build it yourself, or just download the artifact here: https://github.com/actions/setup-go/actions/runs/2051524406

The `dist/` file is what actually gets run when workflows use the action.

brcrista (approved):

PR #228, opened 2022-04-26
Implementation of caching functionality for setup-go action

IvanZosimov:
In the scope of this pull request, the possibility of caching go dependency files and compiler's build outputs was added. Such input parameters as `cache` and `cache-dependency-path` were added.

For now, `cache` input will accept the following values:

true - enable caching for go dependency files and build outputs
false - disable caching for go dependency files and build outputs (default value)

`cache-dependency-path` input is used to specify the path to a dependency file - go.sum

**Description**
Action will try to search go.sum file in the repository root or in any other location which is specified in `cache-dependency-path` input and throw an error if no one is found. The hash of the found file will be used as a cache key ( the same approach as actions/cache recommends). The following key cache will be used: ${{ runner.os }}-go${{ go-version }}-${{ hashFiles('<go.sum-path>') }}

**Examples of use-cases:**

 -  `cache` input only
```yml
steps:
- uses: actions/checkout@v3
- uses: actions/setup-go@v3
  with:
    go-version: '18'
    cache: true
```
 - `cache` along with `cache-dependency-path` 
 ```yml
steps:
- uses: actions/checkout@v3
- uses: actions/setup-go@v3
  with:
    go-version: '18'
    cache: true
    cache-dependency-path: **/go.sum
```

Linked ADR: https://github.com/actions/setup-go/pull/217

> **Note:**
> It's not breaking change because it requires an additional field and nothing changes by default.

**What's done:**
 - Implementation of go dependency files and compiler's build outputs caching
 - Adding unit tests for caching implementation
 - Updating documentation
 - Updating licenses

### CODE REVIEW

**Review by marko-zivic-93 — APPROVED** [review]

File: src/cache-save.ts [hunk 1]
```diff
@@ -0,0 +1,80 @@
...
+  const packageManagerInfo = await getPackageManagerInfo(packageManager);
+
+  const cachePaths = await getCacheDirectoryPath(packageManagerInfo);
+
+  const nonExistingPaths = cachePaths.filter(
+    cachePath => !fs.existsSync(cachePath)
+  );
+
+  if (nonExistingPaths.length === cachePaths.length) {
+    throw new Error(`No cache folders exist on disk`);
```
dmitry-shibanov, line 43: `No cache folders exist on disk` -> 'There are no cache folders on the disk'

src/cache-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+  const obtainedPackageManager = supportedPackageManagers[packageManager];
+
+  return obtainedPackageManager;
+};
+
+export const getCacheDirectoryPath = async (
+  packageManagerInfo: PackageManagerInfo
+) => {
+  let pathList = await Promise.all(
+    packageManagerInfo.cacheFolderCommandList.map(async command =>
```
dmitry-shibanov, line 39: Do we need async here ?

src/cache-save.ts
IvanZosimov: Solved, review again, please.

src/cache-utils.ts
IvanZosimov: Solved, review again, please.

marko-zivic-93 (approved):

dmitry-shibanov (approved):

vsafonkin (approved):

src/cache-restore.ts
```diff
@@ -0,0 +1,62 @@
...
+    : findDependencyFile(packageManagerInfo);
+  const fileHash = await glob.hashFiles(dependencyFilePath);
+
+  if (!fileHash) {
+    throw new Error(
+      'Some specified paths were not resolved, unable to cache dependencies.'
+    );
+  }
+
+  const primaryKey = `${platform}-go${versionSpec}-${fileHash}`;
```
brcrista, line 32: I think it would be great to use a standard format for cache keys.
- For [Node](https://github.com/actions/setup-node/blob/25184c4485a7b08d05f0d0a07d714e3611e40d38/src/cache-restore.ts#L39), it looks like: `node-cache-${platform}-${packageManager}-${fileHash}`
- For [Python / Pip](https://github.com/actions/setup-python/blob/1ce308808af558cfbcb7218da3af37d92e033ad1/src/cache-distributions/pip-cache.ts#L58), it looks like `${this.CACHE_KEY_PREFIX}-${process.env['RUNNER_OS']}-python-${this.pythonVersion}-${this.packageManager}-${hash}`

Now I know those aren't the same, but I think we should follow one or the other so we don't drift so much.  That will help if we're ever able to consolidate the `setup-` caching logic in some package.

.licenses/npm/abort-controller.dep.yml
```diff
@@ -0,0 +1,32 @@
+---
+name: abort-controller
+version: 3.0.0
```
momeyman, line 3: H

.licenses/npm/mime-db.dep.yml
```diff
@@ -0,0 +1,34 @@
+---
```
momeyman, line 1: I

momeyman (approved):

brcrista (approved):

PR #253, opened 2022-08-08
Add support for arm32 go arch

panticmilos:
**Description:**
This PR provides the functionality of specifying architecture for `actions/setup-go`. With this PR support for arm32 will be added as well.

**Check list:**


docs/adrs/0001-architecture.md
```diff
@@ -0,0 +1,16 @@
+## Architecture
+
+You can use any of the [supported operating systems](https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners), and the compatible `architecture` can be selected using `architecture`. Values are `x86`, `x64`, `arm`, `arm64`, `amd64` (not all of the architectures are available on all platforms).
```
marko-zivic-93, line 3: "...and the compatible `architecture` can be selected using `architecture`"
I would switch this with something like:
". Compatible architecture can be selected by specifying a value for 'architecture' input variable."

marko-zivic-93 (approved):

vsafonkin:
I think we should add e2e test with `architecture` input.

marko-zivic-93:
@vsafonkin I think we can't go with e2e tests because of the fact that this is for self hosted arm agents...

PR #265, opened 2022-09-08
fix: add support to pass tokens to GH from GHES

austinsasko:
**Description:**
Removing the logic that limits token authentication on GHES instances (this is a feature that GHES instances can use). Anyone (GHES or not) should be able to pass token as an input value and have it be used.

**Related issue:**
Add link to the related issue.

**Check list:**

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -4,7 +4,7 @@ import * as installer from './installer';
 import * as semver from 'semver';
 import path from 'path';
 import {restoreCache} from './cache-restore';
-import {isGhes, isCacheFeatureAvailable} from './cache-utils';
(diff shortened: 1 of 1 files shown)


brcrista (changes requested):
We found with `setup-python` that this will actually cause an issue with the default value.  See https://github.com/actions/setup-python/pull/437 for more info.

So, you also need to update the default in `action.yml` as in https://github.com/actions/setup-python/pull/443

dmitry-shibanov:
Hello @austinsasko. For now I'm going to close the pull request in favour of https://github.com/actions/setup-go/pull/277

Issue #232, opened 2022-05-04
Add an alias for the previous stable Go release


akshayjshah:
**Description:**
I'd like `setup-go` to support `stable` and `oldstable` as Go version identifiers. `stable` would resolve to the latest patch release of the current stable minor release (so `1.18.x` today), and `oldstable` would resolve to the latest patch release of the _previous_ minor release (so `1.17.x` today). These aliases make it easy for projects to offer the same support policy as the Go language.

**Justification:**
With semver ranges, it's easy to express "use the latest Go release": `^1.18.1` works. However, this is an awkward approach for Go projects:

* Go doesn't use this style of semantic versioning for dependency management, so the syntax is unfamiliar to much of the target audience.
* Many projects mirror [Go's own support policy](https://go.dev/doc/devel/release#policy) and want to test on both `stable` and `oldstable`. Expressing this today requires changing the workflow definition every 6 months.
* Travis CI's `gimme` has supported this for years, starting with https://github.com/travis-ci/gimme/pull/170 from @abhinav.

I'm not attached to the `stable` and `oldstable` names in particular (I think Travis CI wanted `oldstable` because it's similar to a Debian convention) - happy to use whatever you'd like.

**Are you willing to submit a PR?**
Happy to.

akshayjshah:
Any updates, @IvanZosimov?

panticmilos:
Hi @akshayjshah,

Sorry for late response. For now, this feature is not in our plans. :( You can provide a PR with this functionality, we will review it and discuss it further more if that's the case.

akshayjshah:
Sure, happy to put up a PR.

dmitry-shibanov:
Hello everyone. We've released [v3.5.0](https://github.com/actions/setup-go/releases/tag/v3.5.0) and updated the major tag with aliases. Could you please confirm that it works as expected ?

panticmilos:
hi @akshayjshah, we have released this feature, as @dmitry-shibanov mentioned, in the [v3.5.0](https://github.com/actions/setup-go/releases/tag/v3.5.0). 

I will close this issue for now, but when you get a chance to test if there is any issue, please feel free to reopen this issue, or create a new one.

akshayjshah:
Works as expected - thanks for implementing and releasing this! 😻

PR #292, opened 2022-11-20
Add stable and oldstable aliases

panticmilos:
**Description:**
In the scope of this PR, we are adding support for more natural aliases in the Go eco-system, `stable` and `oldstable`. 

**Related issue:**
#232

**Check list:**

### DIFF
--- a/.github/workflows/release-new-action-version.yml
+++ b/.github/workflows/release-new-action-version.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Update the ${{ env.TAG_NAME }} tag
-        uses: actions/publish-action@v0.1.0
+        uses: actions/publish-action@v0.2.1
         with:
           source-tag: ${{ env.TAG_NAME }}
           slack-webhook: ${{ secrets.SLACK_WEBHOOK }}


src/installer.ts
```diff
@@ -8,6 +8,7 @@ import fs from 'fs';
 import os from 'os';
 
 type InstallationType = 'dist' | 'manifest';
+export type StableAliasType = 'stable' | 'oldstable';
```
dmitry-shibanov, line 11: I think it is better to create enums or use a similar approach as for [setup-dotnet](https://github.com/actions/setup-dotnet/blob/main/src/setup-dotnet.ts#L8-L16).

src/main.ts
```diff
@@ -31,6 +31,11 @@ export async function run() {
       let auth = !token ? undefined : `token ${token}`;
 
       const checkLatest = core.getBooleanInput('check-latest');
+
+      if (versionSpec === 'stable' || versionSpec === 'oldstable') {
```
dmitry-shibanov, line 35: I think it is better to create enums or use a similar approach as for [setup-dotnet](https://github.com/actions/setup-dotnet/blob/main/src/setup-dotnet.ts#L8-L16).

dmitry-shibanov:
Could you please update documentation with new syntax description.

src/main.ts
```diff
@@ -143,3 +161,38 @@ function resolveVersionInput(): string {
...
+async function resolveStableVersionInput(
+  versionSpec: string,
+  auth: string | undefined,
+  arch = os.arch(),
+  releases: IToolRelease[]
+): Promise<string> {
+  if (versionSpec === StableReleaseAlias.Stable) {
+    core.info(`Stable version resolved as ${releases[0].version}`);
+
+    return releases[0].version;
```
dmitry-shibanov, line 174: we need to add check for architecture.

src/main.ts
```diff
@@ -143,3 +161,38 @@ function resolveVersionInput(): string {
...
+  } else {
+    const versions = releases.map(
+      release =>
+        `${semver.major(release.version)}.${semver.minor(release.version)}`
+    );
+    const uniqueVersions = Array.from(new Set(versions));
+
+    core.info(`Oldstable version resolved as ${uniqueVersions[1]}`);
```
dmitry-shibanov, line 182: I think it is better to put core.info after `oldstableVersion` to show the full version.

src/main.ts
```diff
@@ -143,3 +148,38 @@ function resolveVersionInput(): string {
 
   return version;
 }
+
+async function resolveStableVersionInput(
```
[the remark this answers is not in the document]
dmitry-shibanov: Could you please move this logic to the `installer.ts`

src/main.ts
```diff
@@ -30,12 +32,29 @@ export async function run() {
...
+        versionSpec === StableReleaseAlias.OldStable
+      ) {
+        versionSpec = await resolveStableVersionInput(
+          versionSpec,
+          auth,
+          arch,
+          releases,
+          checkLatest
+        );
+      }
```
dmitry-shibanov, line 50: Could you please move this logic to the `installer.ts`

src/installer.ts
```diff
@@ -174,21 +183,27 @@ export async function extractGoArchive(archivePath: string): Promise<string> {
...
-    'go-versions',
-    auth,
-    'main'
-  );
+  releases = releases ? releases : await getAllManifestReleases(auth);
+
   core.info(`matching ${versionSpec}...`);
-  const rel = await tc.findFromManifest(versionSpec, stable, releases, arch);
+
+  let rel = await tc.findFromManifest(versionSpec, stable, releases, arch);
```
dmitry-shibanov, line 206: I think we can keep it const

src/installer.ts
```diff
@@ -174,21 +183,27 @@ export async function extractGoArchive(archivePath: string): Promise<string> {
...
+  releases?: tc.IToolRelease[] | undefined
 ): Promise<IGoVersionInfo | null> {
   let info: IGoVersionInfo | null = null;
-  const releases = await tc.getManifestFromRepo(
-    'actions',
-    'go-versions',
-    auth,
-    'main'
-  );
+  releases = releases ? releases : await getAllManifestReleases(auth);
```
dmitry-shibanov, line 202: Could you please change it to this snippet:
```
  if (!manifest) {
    core.debug('No manifest cached');
    manifest = await getManifest(auth);
  }
```

src/installer.ts
```diff
@@ -174,21 +183,27 @@ export async function extractGoArchive(archivePath: string): Promise<string> {
   return extPath;
 }
 
+export async function getAllManifestReleases(auth: string | undefined) {
+  return tc.getManifestFromRepo('actions', 'go-versions', auth, 'main');
+}
+
+export async function getAllToolCacheReleases(arch = os.arch()) {
```
dmitry-shibanov, line 190: Could you please rename it to `getAllToolCacheVersions`

src/installer.ts
```diff
@@ -174,21 +183,27 @@ export async function extractGoArchive(archivePath: string): Promise<string> {
   return extPath;
 }
 
+export async function getAllManifestReleases(auth: string | undefined) {
```
dmitry-shibanov, line 186: Could you please rename it to `getManifest`

src/main.ts
```diff
@@ -4,18 +4,20 @@ import * as installer from './installer';
...
+import {StableReleaseAlias} from './utils';
 
 export async function run() {
   try {
     //
     // versionSpec is optional.  If supplied, install / use from the tool cache
     // If not supplied then problem matchers will still be setup.  Useful for self-hosted.
     //
-    const versionSpec = resolveVersionInput();
+    let versionSpec = resolveVersionInput();
```
marko-zivic-93, line 20: When you move the logic down that Dmitry suggested to installer.ts, you will no longer need this variable to be defined as let, you will be able to use const again.

marko-zivic-93 (approved):

README.md
```diff
@@ -16,6 +16,7 @@ The V3 edition of the action offers:
 - Proxy support
 - Check latest version
 - Caching packages dependencies
+- stable and olstable aliases
```
brcrista, line 19:
```suggestion
- stable and oldstable aliases
```

.github/workflows/versions.yml
```diff
@@ -12,6 +12,72 @@ on:
...
+
+  stable-no-check-latest:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup Go and check latest
```
brcrista, line 57: `check-latest` isn't set for this step and the step in `oldstable-no-check-latest`

README.md
```diff
@@ -95,6 +96,31 @@ steps:
       check-latest: true
   - run: go run hello.go
 ```
+
+## Using stable/oldstable aliases
+
+Given the fact that Go doesn't use semver syntax ranges, and that many Go projects want to run tests based on `stable` and `oldstable` aliases, these aliases are introduced in the action as possible values of the input `go-version`. When alias `stable` is provided, and `check-latest` input is set to `true` action will get the latest stable version from the [`go-versions`](https://github.com/actions/go-versions/blob/main/versions-manifest.json) repository manifest. When `check-latest` is not provided, or set to false, the action will resolve `stable` as the most recent present version from the runners tool cache directory. In case of the `oldstable` alias, if current release is 1.19.x, action will resolve version as 1.18.x, where x is the latest patch release. The `check-latest` rules described above will apply for `oldstable` as well.
```
brcrista, line 102: Can we have `stable` and `oldstable` imply `check-latest`?

brcrista (changes requested):
Can we have `stable` and `oldstable` imply `check-latest`?

That will simplify the user-facing behavior, and I suspect it aligns with what people want when they ask for `stable`.  It will also reduce the number of cases we have to cover in our tests.

brcrista (approved):

e-korolevskii (approved):

src/main.ts
```diff
@@ -30,14 +31,24 @@ export async function run() {
...
+        arch,
+        manifest
       );
 
+      if (IS_WINDOWS) {
+        versionSpec = installDir.split('\\').reverse()[1];
+      } else {
+        versionSpec = installDir.split('/').reverse()[1];
+      }
+
```
dmitry-shibanov, line 51: Could you please change it to `const version = path.basename(path.dirname(installDir))`

src/utils.ts
```diff
@@ -0,0 +1,6 @@
+export enum StableReleaseAlias {
+  Stable = 'stable',
+  OldStable = 'oldstable'
+}
+
+export const IS_WINDOWS = process.platform === 'win32';
```
dmitry-shibanov, line 6: I think we can remove it.

src/main.ts
```diff
@@ -4,18 +4,19 @@ import * as installer from './installer';
...
+import {IS_WINDOWS} from './utils';
 
 export async function run() {
   try {
     //
     // versionSpec is optional.  If supplied, install / use from the tool cache
     // If not supplied then problem matchers will still be setup.  Useful for self-hosted.
     //
-    const versionSpec = resolveVersionInput();
+    let versionSpec = resolveVersionInput();
```
dmitry-shibanov, line 19: Could you please revert changes and keep it const.

src/installer.ts
```diff
@@ -114,14 +129,21 @@ export async function getGo(
   return downloadPath;
 }
 
-async function resolveVersionFromManifest(
+export async function resolveVersionFromManifest(
```
dmitry-shibanov, line 132: I think we can remove export.

.github/workflows/versions.yml
```diff
@@ -12,6 +12,38 @@ on:
...
+        uses: ./
+        with:
+          go-version: oldstable
+          architecture: x64
```
dmitry-shibanov, line 43: I think we can remove architecture.

Issue #297, opened 2022-12-04
Suggestion: Use early return pattern to avoid nested conditions


jongwooo:
## Description

Return early is the way of writing functions or methods so that the expected positive result is returned at the end of the function and the rest of the code terminates the execution (by returning or throwing an exception) when conditions are not met.

See [actions/cache#1012](https://github.com/actions/cache/issues/1012)

In `cache-utils.ts`:

**AS-IS**

```typescript
export function isCacheFeatureAvailable(): boolean {
  if (!cache.isFeatureAvailable()) {
    if (isGhes()) {
      throw new Error(
        'Cache action is only supported on GHES version >= 3.5...'
      );
    } else {
      core.warning(
        'The runner was not able to contact...'
      );
    }

    return false;
  }

  return true;
}
```

**TO-BE**

```typescript
export function isCacheFeatureAvailable(): boolean {
  if (cache.isFeatureAvailable()) {
    return true;
  }

  if (isGhes()) {
    throw new Error(
      'Cache action is only supported on GHES version >= 3.5...'
    );
  }

  core.warning(
    'The runner was not able to contact...'
  );
  return false;
}
```

e-korolevskii:
Hi @jongwooo 

We will investigate this issue and come back to you.

e-korolevskii:
Hi @jongwooo!

Thanks for your contribution!
The changes will be available in the next release. I am closing this issue for now!

PR #302, opened 2022-12-12
refactor: Use early return pattern to avoid nested conditions

jongwooo:
Signed-off-by: jongwooo <jongwooo.han@gmail.com>
Issue #297 

## Description

Return early is the way of writing functions or methods so that the expected positive result is returned at the end of the function and the rest of the code terminates the execution (by returning or throwing an exception) when conditions are not met.

See [actions/cache#1012](https://github.com/actions/cache/issues/1012), [actions/setup-node#639](https://github.com/actions/setup-node/pull/639)

In `cache-utils.ts`:

**AS-IS**

```typescript
export function isCacheFeatureAvailable(): boolean {
  if (!cache.isFeatureAvailable()) {
    if (isGhes()) {
      throw new Error(
        'Cache action is only supported on GHES version >= 3.5...'
      );
    } else {
      core.warning(
        'The runner was not able to contact...'
      );
    }

    return false;
  }

  return true;
}
```

**TO-BE**

```typescript
export function isCacheFeatureAvailable(): boolean {
  if (cache.isFeatureAvailable()) {
    return true;
  }

  if (isGhes()) {
    throw new Error(
      'Cache action is only supported on GHES version >= 3.5...'
    );
  }

  core.warning(
    'The runner was not able to contact...'
  );
  return false;
}
```

**Check list:**

### DIFF
--- a/__tests__/cache-utils.test.ts
+++ b/__tests__/cache-utils.test.ts
@@ -162,7 +162,7 @@ describe('isCacheFeatureAvailable', () => {
     expect(functionResult).toBeFalsy();
   });
 
-  it('should throw when cache feature is unavailable and GHES is used', () => {
+  it('should warn when cache feature is unavailable and GHES is used', () => {
     //Arrange
     isFeatureAvailableSpy.mockImplementation(() => {
       return false;
@@ -170,10 +170,11 @@ describe('isCacheFeatureAvailable', () => {
 
     process.env['GITHUB_SERVER_URL'] = 'https://nongithub.com';
 
-    let errorMessage =
+    let warningMessage =
       'Cache action is only supported on GHES version >= 3.5. If you are on version >=3.5 Please check with GHES admin if Actions cache service is enabled or not.';
 
     //Act + Assert
-    expect(() => cacheUtils.isCacheFeatureAvailable()).toThrow(errorMessage);
+    expect(cacheUtils.isCacheFeatureAvailable()).toBeFalsy();
+    expect(warningSpy).toHaveBeenCalledWith(warningMessage);
   });
 });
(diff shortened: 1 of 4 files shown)


panticmilos:
hi @jongwooo again :D thank you for the PR. We will take a look at it

src/cache-utils.ts
```diff
@@ -57,19 +57,18 @@ export function isGhes(): boolean {
...
-        'The runner was not able to contact the cache service. Caching will be skipped'
-      );
-    }
+  if (cache.isFeatureAvailable()) {
+    return true;
+  }
 
-    return false;
+  if (isGhes()) {
+    throw new Error(
```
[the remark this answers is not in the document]
jongwooo: @e-korolevskii Thank you for your review! I also updated the testcode.

e-korolevskii (approved):

marko-zivic-93 (approved):

__tests__/cache-utils.test.ts
```diff
@@ -174,6 +174,7 @@ describe('isCacheFeatureAvailable', () => {
       'Cache action is only supported on GHES version >= 3.5. If you are on version >=3.5 Please check with GHES admin if Actions cache service is enabled or not.';
 
     //Act + Assert
-    expect(() => cacheUtils.isCacheFeatureAvailable()).toThrow(errorMessage);
+    expect(cacheUtils.isCacheFeatureAvailable()).toBeFalsy();
+    expect(warningSpy).toHaveBeenCalledWith(errorMessage);
```
MaksimZhukov, line 178: Please rename the `errorMessage` variable to `warningMessage`. 
I think we also need to rename this test case to `should warn when cache feature is unavailable and GHES is used`

jongwooo: @MaksimZhukov Thank you for your review. I pushed a commit.

MaksimZhukov (approved):

dmitry-shibanov (approved):

PR #321, opened 2023-01-27
Enable caching by default

dsame:
**Description:**
Change default value for `cache` input

**Related issue:**
[Add link to the related issue.](https://github.com/actions/runner-images-internal/issues/4764)

**Check list:**


src/package-managers.ts
```diff
@@ -13,3 +13,5 @@ export const supportedPackageManagers: SupportedPackageManagers = {
     cacheFolderCommandList: ['go env GOMODCACHE', 'go env GOCACHE']
   }
 };
+
+export const getCurrentPackageManager = (): string => 'default';
```
e-korolevskii, line 17:
```suggestion
export const getDefaultPackageManager = (): string => 'default';
```

action.yml
```diff
@@ -14,7 +14,7 @@ inputs:
     default: ${{ github.server_url == 'https://github.com' && github.token || '' }}
   cache:
     description: Used to specify whether caching is needed. Set to true, if you'd like to enable caching.
-    default: false
+    required: false
```
e-korolevskii, line 17:
```suggestion
    default: true
    required: false
```

src/cache-utils.ts
```diff
@@ -73,3 +78,16 @@ export function isCacheFeatureAvailable(): boolean {
   );
   return false;
 }
+
+export async function isCacheEnabled() {
```
e-korolevskii, line 82: We might split it into two functions: 
`isCacheEnabled` and `isCachePathsExists`

dsame:
The PR is aborted due to https://github.com/actions/setup-go/pull/332

Issue #324, opened 2023-01-31
Allow setting cache key for multiple targets (same go.mod)


hermanbanken:
**Description:**
Any larger project with multiple cmd might build multiple binaries with distinct sets of dependencies. As #259 already indicated, those workflows will fight for the cache. Given workflow A & B, if any of those two finishes and is cached, any subsequent build reuses the same cache. Even if the build of B has a completely different `go-build` cache state. 

**Justification:**
This is specifically a problem, when  building both (tiny) plugins, and large server binaries (both in different pipelines). This requires a monorepo, because the plugin must use the exact same module versions, so the go.mod/go.sum files are the same, therefore `cache-dependency-path` **is not an option.**

Both should use the same `go env GOMODCACHE` (`go/pkg/mod`) but different `go env GOCACHE` (`go-build`).

**Are you willing to submit a PR?**
Should I create a PR for an optional field that can be set to be included in the cache key?

I think something like the following would suffice:

```yaml
# workflow job A
steps:
  - uses: actions/checkout@v3
  - uses: actions/setup-go@v3
    with:
      go-version: '1.19'
      cache: true
      cache-key-suffix: toolA
  - run: go run cmd/toolA/

# workflow job B
steps:
  - uses: actions/checkout@v3
  - uses: actions/setup-go@v3
    with:
      go-version: '1.19'
      cache: true
      cache-key-suffix: toolB
  - run: go run cmd/toolB/
 
```

e-korolevskii:
Hello @hermanbanken. Thank you for your report. We'll investigate the feature request.

dmitry-shibanov:
Hello @hermanbanken. I'm going to close the issue because according to the [ADR](https://github.com/actions/setup-go/blob/main/docs/adrs/0000-caching-dependencies.md#goals--anti-goals) the goal of caching is to provide basic cases for its usage. If you need more flexibility you should use [actions/cache](https://github.com/actions/cache).

Issue #259, opened 2022-09-01
Multiple workflows will fight for the cache


eddycharly:
We have multiple workflows running with the same repo.
They seem to keep fighting for the same cache key.
Is there a way to have a different key per workflow ? Shall we do this or is it a bad idea ?

e-korolevskii:
Hey @eddycharly!

Thanks for this issue. You are correct that the workflows will use a same cache key.
Could you explain in more detail why you need different keys? Preferably with examples of runs.

eddycharly:
@e-korolevskii i am caching go dependencies. But those dependencies vary depending on the makefile targets invoked.
We use `go install` to install various tools.

e-korolevskii:
Hi @eddycharly 

In your case, you can try using the [cache-dependency-path input](https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs). Since this is expected behavior, I'm going to close this issue.

PR #325, opened 2023-02-01
feature: allow diversification using cache-key-suffix

hermanbanken:
**Description:**
Any larger project with multiple cmd might build multiple binaries with distinct sets of dependencies. As https://github.com/actions/setup-go/issues/259 already indicated, those workflows will fight for the cache. Given workflow A & B, if any of those two finishes and is cached, any subsequent build reuses the same cache. Even if the build of B has a completely different `go-build` cache state. This means that for one of the workflows, the build will always be slow.

This is specifically a problem, when building both (tiny) plugins, and large server binaries (both in different pipelines). This requires a monorepo, because the plugin must use the exact same module versions, so the go.mod/go.sum files are the same, therefore `cache-dependency-path` **is not an option**.

Both should use the same `go env GOMODCACHE` (`go/pkg/mod`) but different `go env GOCACHE` (`go-build`).

Usage:

```yaml
# workflow job A
steps:
  - uses: actions/checkout@v3
  - uses: actions/setup-go@v3
    with:
      go-version: '1.19'
      cache: true
      cache-key-suffix: toolA
  - run: go run cmd/toolA/

# workflow job B
steps:
  - uses: actions/checkout@v3
  - uses: actions/setup-go@v3
    with:
      go-version: '1.19'
      cache: true
      cache-key-suffix: toolB
  - run: go run cmd/toolB/
```

**Related issue:**
Implements #324, #259

**Check list:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,33 @@ steps:
       cache-dependency-path: subdir/go.sum
   - run: go run hello.go
   ```
+
+**Caching for different build profiles**
+Suppose you have multiple builds that depend on different dependencies, spread over multiple workflows or jobs. 
+If the same go.sum file is used, they will fight for the cache. Set `cache-key-suffix` to distinguish these caches.
+
+```yaml
+steps:
+  - uses: actions/checkout@v3
+  - uses: actions/setup-go@v3
+    with:
+      go-version: '1.17'
+      check-latest: true
+      cache: true
+      cache-key-suffix: hello
+  - run: go run hello.go
+
+# another workflow job
+steps:
+  - uses: actions/checkout@v3
+  - uses: actions/setup-go@v3
+    with:
+      go-version: '1.17'
+      check-latest: true
+      cache: true
+      cache-key-suffix: server
+  - run: go build -o /server ./cmd/
+  ```
 ## Getting go version from the go.mod file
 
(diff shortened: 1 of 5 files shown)


dsame (approved):

dmitry-shibanov:
I'm going to close the pull request because the relevant issue is closed.

hermanbanken:
My fix what a little broader than the issue. In the case of the issue, using a different `cache-dependency-path` was an option.

In the case of multiple sub-commands in the same go module, different `cache-dependency-path` is not possible. go.mod/go.sum files are the same for the entire go repository.

I don't understand how people use this setup-go action for more complex repositories without running into this issue. Maybe @dmitry-shibanov or @e-korolevskii can explain what I'm missing, or why they think this is not needed.

hermanbanken:
The alternative is self-hosting such an composite action:

```yaml
name: Setup Go
description: Install go with optimal caching

inputs:
  key: # id of input
    description: 'How to identify the cache unique per step; example "job:lint:go"'
    required: true

runs:
  using: "composite"
  steps:
    - name: Cache go
      uses: actions/cache@v3
      with:
        key: ${{ inputs.key }}
        path: | # what setup-go caches too (runs `go env GOMODCACHE` and `go env GOCACHE` to compute this, but it is static)
          /home/runner/go/pkg/mod
          /home/runner/.cache/go-build
    - name: Setup go
      uses: actions/setup-go@v3
      with:
        go-version-file: go.mod
        cache: false
```

PR #351, opened 2023-03-13
Add Go bin if go-version input is empty

dmitry-shibanov:
**Description:**
In scope of this pull request we update logic to add Go bin even if go-version input is empty. 

**Related issue:**
https://github.com/actions/setup-go/issues/49

**Check list:**

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -50,6 +50,8 @@ export async function run() {
         core.info('Setting GOROOT for Go version < 1.9');
         core.exportVariable('GOROOT', installDir);
       }
+      
+      core.info(`Successfully set up Go version ${versionSpec}`);
     } else {
       core.info(
         '[warning]go-version input was not specified. The action will try to use pre-installed version.'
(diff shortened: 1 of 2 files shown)


panticmilos (approved):

src/main.ts
```diff
@@ -50,12 +50,12 @@ export async function run() {
...
       }
-
-      const added = await addBinToPath();
-      core.debug(`add bin ${added}`);
-      core.info(`Successfully set up Go version ${versionSpec}`);
     }
 
+    const added = await addBinToPath();
+    core.debug(`add bin ${added}`);
+    core.info(`Successfully set up Go version ${versionSpec}`);
```
MaksimZhukov, line 57: I think we should leave `core.info` in the `if` statement because it can be empty if no version is specified.

IvanZosimov: But can we say that we have successfully installed go before the `go/bin` was added to the path ? I think we have to wrap 57 line in condition depends on version spec is set or not. If set - we do installed the version, If not we just added precached version to the path - no need to log anything.
```suggestion
    if (versionSpec) {
      core.info(`Successfully set up Go version ${versionSpec}`)
    };
```

Issue #375, opened 2023-05-19
feat: Utilize the`.tool-versions` file to set expected Go version


plukevdh:
**Description:**
It would be useful to utilize the [`.tool-versions` file](https://asdf-vm.com/manage/configuration.html#tool-versions) to set the expect version of Go for the action runtime _in addition_ to being able to set via the parameters.

**Justification:**
Because an action may be used across a variety of steps, it can be burdensome to track the Go version via the parameter in every place used. Several other `setup-<lang>` projects are also utilizing or making progress towards using the `.tool-versions` file for version locking.

- Ruby [currently utilizes](https://github.com/ruby/setup-ruby#supported-version-syntax)
- Python has a [PR open](https://github.com/actions/setup-python/pull/588) to add this functionality
- Node [uses it](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file)

**Are you willing to submit a PR?**
💯 if this feels like a thing that would be useful to the community and in keeping with the goal of this project.

IvanZosimov:
Hi, @plukevdh 👋 Thanks for the feature request, we will take a look and get back to you with updates.

plukevdh:
@IvanZosimov thanks! Note that I also opened a PR for this feature as well with some initial ideas: https://github.com/actions/setup-go/pull/376

steebchen:
This would be awesome to have, the setup-node action also supports this 👍 


> ```
> - uses: actions/setup-node@v3
>   with:
>     # ...
>     # File containing the version Spec of the version to use.  Examples: .nvmrc, .node-version, .tool-versions.
>     # If node-version and node-version-file are both provided the action will use version from node-version. 
>     node-version-file: ''
> ```
> 
> https://github.com/actions/setup-node#usage

dsame:
Hello @plukevdh , thank you for your contribution, but i would ask you to review this PR as well https://github.com/actions/setup-go/pull/402 - do you see any drawbacks of using the existing `go-version-file` input instead of introducing new one `tool-versions-file` https://github.com/actions/setup-go/pull/376/files#diff-b97cd7a3fa566070832c220ba3fab8e4f60694550f6fe74d6024a20486165de8R148 ?

dsame:
Hello @plukevdh , thank you for assistance to make a choice between #402 and #376, but i'll keep this issue open till #402 will have been merged

priya-kinthali:
Hello👋,
[...]

PR #376, opened 2023-05-19
Parse .tool-versions file to determine Go version

plukevdh:
**Description:**
Allow `setup-go` to leverage the `.tool-versions` file for determining the Go version desired. As noted in the linked issue below
- Ruby [currently utilizes](https://github.com/ruby/setup-ruby#supported-version-syntax)
- Python has a https://github.com/actions/setup-python/pull/588 to add this functionality
- Node [uses it](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file)

**Related issue:**
#375

**Check list:**

**Please discuss:**
- This change currently makes .tool-versions the backup/fallback option if no other version spec is given. Is that desired?
- ASDF/`.tool-versions` requires a fairly verboase version pattern which may or may not meet current version expectations. I see that there are a few cannonican version parsing strategies in the codebase here. Should the version be passed through those parsers or will the `.tool-versions` version spec be adequate?
- I'm also unsure if the current default to `.tool-versions` path is going to be at project root. This is honestly my first attempt at contributing to a GH action codebase. I'd appreciate any "best practice" guidance available
- ASDF also accepts the following version formats:
  - `ref:v1.0.2-a` or `ref:39cb398vb39` - tag/commit/branch to download from github and compile
  - `path:~/src/elixir` - a path to custom compiled version of a tool to use. For use by language developers and such.
  - `system` - this keyword causes asdf to passthrough to the version of the tool on the system that is not managed by asdf.

  I'm not sure if any of these make sense to support given that this action currently only otherwise supports semver versions of the language.

### DIFF
--- a/.github/workflows/versions.yml
+++ b/.github/workflows/versions.yml
@@ -139,7 +139,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, windows-latest, macos-latest ]
+        os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
       - uses: actions/checkout@v3
       - name: Setup Go and check latest
--- a/.github/workflows/versions.yml
+++ b/.github/workflows/versions.yml
@@ -134,6 +134,22 @@ jobs:
         run: __tests__/verify-go.sh 1.19
         shell: bash
 
+  tool-versions-file:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest, windows-latest, macos-latest ]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup Go and check latest
+        uses: ./
+        with:
+          tool-versions-file: __tests__/data/.tool-versions
+      - name: verify go
+        run: __tests__/verify-go.sh 1.16.3
+        shell: bash
+
   setup-versions-from-manifest:
     name: Setup ${{ matrix.go }} ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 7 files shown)


songu21:
d

songu21:
dddjdk

bflad:
Is there a particular reason why ASDF handling is being introduced into various actions rather than introducing an ASDF-specific action which parses and outputs per-tool version information? e.g. In psuedo-workflow configuration:

```yaml
- uses: actions/checkout
- uses: xxx/read-asdf
  id: asdf
- uses: actions/setup-go
  with:
    go-version: steps.asdf.outputs.go
```

This enables downstream actions to remain generic within their domain, rather than needing to understand and maintain a different tooling ecosystem.

plukevdh:
@bflad A good question. I could see a version parser being added as a separate action. My argument is that it's being added to a lot of other actions across the language ecosystem so why not add here as well. ASDF does provide an action for installing all the things, but it doesn't (and arguably shouldn't) take into account all the potential dependency caching mechanisms most of these language-specific libraries add. I could definitely look to add a tool version output parser to the ASDF action as well, but I'd still see value in this addition. I believe ASDF is attempting to put forward a tool agnostic version spec file (as opposed to a `.ruby-version` or `.node-version` file) which could be adopted more globally.

dsame:
relates to https://github.com/actions/setup-go/issues/375

dsame:
@plukevdh do you think the issue can to be solved with https://github.com/actions/setup-go/pull/376 ?

plukevdh:
@dsame can you rephrase that question? Are you asking if #402 solve this PR as well? I think it would, though I wouldn't be opposed to seeing more tests to demonstrate the solve.

plukevdh:
Closing in favor of #402

9777410929:
> Closing in favor of #402 

hi

Issue #380, opened 2023-06-02
Using `go-version: 1.20` breaks the build


metalim:
**Description:**
Using `go-version: 1.20` breaks the build, as it uses Go 1.2 instead of 1.20

**Action version:**
v3 and v4

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**
1.20

**Repro steps:**  
* Create Go project with `go.mod` in it, with proper module name (in my case: `github.com/metalim/gotest`)
* Add test file that imports that module
* Confirm that `go test ./...` passes locally
* Setup GH action with Go template
* Change `go-version` for `setup-go` from 1.19 to 1.20
* Observe build failing. Example: https://github.com/metalim/gotest/actions

**Expected behavior:**
Should use Go 1.20 or 1.20.4

**Actual behavior:**
Uses Go 1.2

metalim:
As this was reported before, I suggest editing the Go template of GH action and using quoted version there, so it does not make confusion again

marko-zivic-93:
Hello @metalim 
Thank you for the heads up! We will investigate this issue and come back to you as soon as we have some results. 🙂

nfuden:
A workaround is to use the go-version-file option as 1.20 will work in that setup

gaby:
Fixed via #382

dmitry-shibanov:
Hello everyone. I'm going to close the pull request because we added a note in scope of this pull request: https://github.com/actions/setup-go/pull/382

metalim:
That might not help, and close was too early.
The cause of this issue originates from template provided by github, when you select template for Go Action:
<img width="907" alt="image" src="https://github.com/actions/setup-go/assets/230431/f354b906-20a1-4e1c-b5f3-40c9e40f3230">

<img width="715" alt="image" src="https://github.com/actions/setup-go/assets/230431/1a3df061-dfc2-485a-bd28-df7709c45791">

README is not available at that point, but value already has incorrect format.
Requesting re-open

PR #382, opened 2023-06-05
Add note about YAML parsing versions

dmitry-shibanov:
**Description:**
In scope of this pull request we add note about YAML parsing versions like `1.20`.

**Related issue:**
https://github.com/actions/setup-go/issues/380

**Check list:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,13 +58,13 @@ steps:
   - run: go version
 ```
 
-> **_NOTE:_** It is better to enclose the version in single quotation mark, due to YAML parsing. When the input is specified like this  
-> ```
->  1.20 or >=1.17.0
+> **Note**: Due to the peculiarities of YAML parsing, it is recommended to wrap the version in single quotation marks:
+> 
+> ```yaml
+>   go-version: '1.20'
 >  ```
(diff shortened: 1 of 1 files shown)


dsame (approved):

README.md
```diff
@@ -58,6 +58,13 @@ steps:
   - run: go version
 ```
 
+> **_NOTE:_** It is better to enclose the version in single quotation mark, due to YAML parsing. When the input is specified like this  
+> ```
+>  1.20 or >=1.17.0
+>  ```
+> 1.20 is parsed as a number and it is trimmed to 1.2 that is an expected behavior for numbers. You should specify input with quotes to treat 
+> it as a string.
+
```
IvanZosimov, line 67:
```suggestion
> **Note**: Due to the peculiarities of YAML parsing, it is recommended to wrap the version in single quotation marks:
> 
> ```yaml
>   go-version: '1.20'
>  ```
>  
> The recommendation is based on the YAML parser's behavior, which interprets non-wrapped values as numbers and, in the case of version 1.20, trims it down to 1.2, which may not be very obvious.
```
**Nit:** I'd change the `**_NOTE:_**` to `**Note**:` to use default GitHub visualization of this word. Also I rephrased the sentences a little.

IvanZosimov (approved):

Issue #368, opened 2023-04-10
cache key do not include linux distro name and version


kolyshkin:
**Description:**

setup-go@v4 added `/home/runner/.cache/go-build` caching, which does not take the OS version into account, as a result, cache saved from Ubuntu 22.04 can be used on Ubuntu 20.04, resulting in cryptic linker errors. Apparently, this issue is only happening with Go 1.19.x (I guess they've worked around that in Go 1.20, but am too lazy to take a look).

Found it by experimenting with adding Ubuntu 22.04 to github.com/opencontainers/runc CI matrix. Meaning, the tests are now run on BOTH Ubuntu 20.04 and 22.04, using both Go 1.19.x and 1.20.x. IOW, here's the test matrix, simplified:

```yaml
jobs:
  test:
    strategy:
      matrix:
        os: [ubuntu-20.04, ubuntu-22.04]
        go-version: [1.19.x, 1.20.x]
    runs-on: ${{ matrix.os }}
```

Once I managed to get it running, I started to see errors like these on Ubuntu 20.04:

```
/tmp/go-build2163232479/b244/fs.test: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /tmp/go-build2163232479/b244/fs.test)
402
/tmp/go-build2163232479/b244/fs.test: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /tmp/go-build2163232479/b244/fs.test)
403
FAIL	github.com/opencontainers/runc/libcontainer/cgroups/fs	0.001s
```

```
# /home/runner/work/runc/runc/tests/integration/../../contrib/cmd/recvtty/recvtty: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /home/runner/work/runc/runc/tests/integration/../../contrib/cmd/recvtty/recvtty)
646
# /home/runner/work/runc/runc/tests/integration/../../contrib/cmd/recvtty/recvtty: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /home/runner/work/runc/runc/tests/integration/../../contrib/cmd/recvtty/recvtty)
```

By looking into setup-go logs for Ubuntu-20.04 + Go 1.19.x, I see:

```
Successfully set up Go version 1.19.x
/opt/hostedtoolcache/go/1.19.7/x64/bin/go env GOMODCACHE
[...]

PR #383, opened 2023-06-06
Add imageOS to primaryKey

dsame:
**Description:**
Cache key now contains the os version in order to avoid conflicts between the binaries built against the unexpected libraries

TBD: the original issue explicitly mentions Linux as the os having issue. Probably the solution should be limited to Linux only.

**Related issue:**
[link to the related issue.](https://github.com/actions/setup-go/issues/368)

**Check list:**

### DIFF
--- a/src/cache-restore.ts
+++ b/src/cache-restore.ts
@@ -29,8 +29,9 @@ export const restoreCache = async (
     );
   }
 
-  const imageOS = process.env.ImageOS || 'self-hosted';
-  const primaryKey = `setup-go-${platform}-${imageOS}-go-${versionSpec}-${fileHash}`;
+  const linuxVersion =
+    process.env.RUNNER_OS === 'Linux' ? `${process.env.ImageOS}-` : '';
+  const primaryKey = `setup-go-${platform}-${linuxVersion}go-${versionSpec}-${fileHash}`;
   core.debug(`primary key is ${primaryKey}`);
 
   core.saveState(State.CachePrimaryKey, primaryKey);
--- a/src/cache-restore.ts
+++ b/src/cache-restore.ts
@@ -29,7 +29,8 @@ export const restoreCache = async (
     );
   }
 
-  const primaryKey = `setup-go-${platform}-go-${versionSpec}-${fileHash}`;
+  const imageOS = process.env.ImageOS || 'self-hosted';
+  const primaryKey = `setup-go-${platform}-${imageOS}-go-${versionSpec}-${fileHash}`;
   core.debug(`primary key is ${primaryKey}`);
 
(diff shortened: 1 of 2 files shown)


marko-zivic-93 (approved):

src/cache-restore.ts
```diff
@@ -29,7 +29,8 @@ export const restoreCache = async (
     );
   }
 
-  const primaryKey = `setup-go-${platform}-go-${versionSpec}-${fileHash}`;
+  const imageOS = process.env.ImageOS || 'self-hosted';
+  const primaryKey = `setup-go-${platform}-${imageOS}-go-${versionSpec}-${fileHash}`;
```
dmitry-shibanov, line 33: I agree with you. I think the solution should be limited only of Linux. Could you please change it.

src/cache-restore.ts
```diff
@@ -29,7 +29,9 @@ export const restoreCache = async (
     );
   }
 
-  const primaryKey = `setup-go-${platform}-go-${versionSpec}-${fileHash}`;
+  const linuxVersion =
+    (process.env.RUNNER_OS === 'Linux' && `${process.env.ImageOS}-`) || '-';
```
dmitry-shibanov, line 33: You need to change `${process.env.ImageOS}-` because it won't be falsy value

dmitry-shibanov: We can keep it if we're sure that ImageOS persist with RUNNER_OS

dsame: I want to check for "${process.env.ImageOS}-" is not false for sake of self-hosted environments. The probability of RUNNER_OS is set and ImageOS is not set is low but still above zero.

src/cache-restore.ts
```diff
@@ -29,7 +29,9 @@ export const restoreCache = async (
     );
   }
 
-  const primaryKey = `setup-go-${platform}-go-${versionSpec}-${fileHash}`;
+  const linuxVersion =
+    process.env.RUNNER_OS === 'Linux' ? `${process.env.ImageOS}-` : '-';
+  const primaryKey = `setup-go-${platform}-${linuxVersion}go-${versionSpec}-${fileHash}`;
```
dusan-trickovic, line 34: LGTM! I do have a question about this line though. It's a small remark, but if the condition in `linuxVersion` is `false`, would there not be a double hyphen (--) inside the key? Again, just a small observation, but worth asking I suppose

dusan-trickovic (approved):

src/cache-restore.ts
dsame: fixed

nikolai-laevskii (approved):

marko-zivic-93 (approved):

Issue #240, opened 2022-06-24
Install on Windows is very slow


G-Rath:
**Description:**
Installing Go on Windows seems to consistently take at least 30+ seconds to install, usually 1 minute and sometimes more - it's at least consistently 30 seconds longer than on other OSs (which typically take 10-20 seconds).

This consistently represents ~3/4th of the CI run time for [`osv-detector`](https://github.com/G-Rath/osv-detector) - the test suite takes only ~20 seconds (on Windows).

![image](https://user-images.githubusercontent.com/3151613/175659615-72af83e8-d05a-40a9-852c-8d430af2cf66.png)

![image](https://user-images.githubusercontent.com/3151613/175659635-c8e74b7b-a939-4f43-9106-5bc713d77cff.png)

https://github.com/G-Rath/osv-detector/runs/7033093878?check_suite_focus=true

This one took 3 minutes:

https://github.com/G-Rath/osv-detector/runs/7033278935?check_suite_focus=true

**Action version:**
v3

**Platform:**
- [ ] Ubuntu
- [ ] macOS
- [x] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**

1.17.9

**Repro steps:**  
https://github.com/G-Rath/osv-detector/blob/main/.github/workflows/checks.yml

https://github.com/G-Rath/osv-detector/actions/workflows/checks.yml
https://github.com/G-Rath/osv-detector/runs/7033093878?check_suite_focus=true
https://github.com/G-Rath/osv-detector/runs/7034205194?check_suite_focus=true
https://github.com/G-Rath/osv-detector/runs/7033293623?check_suite_focus=true
https://github.com/G-Rath/osv-detector/runs/7033278935?check_suite_focus=true

**Expected behavior:**
The install time to be below 30 seconds on average, ideally never going over a minute.

**Actual behavior:**
The install time is consistently above 30 seconds, often even above 1 minute and sometimes as long as 3 minutes.

dmitry-shibanov:
Hello @G-Rath. Thank you for your report. We'll investigate the issue.

panticmilos:
Hi @G-Rath,

We need to replace extracting method in order to fix this issue. I will let you know when that PR is ready.

ValentinTrinque:
Is this something that is still planned ?

On one of my setup:
- Linux takes ~20s
- MacOS is a bit slower and takes ~1m
- But Windows takes ~3m20s

[...]

PR #393, opened 2023-07-11
Fix Install on Windows is very slow

dsame:
**Description:**

In order to avoid slow IO of C: drive the go executables are cached to D: drive instead of C: drive and create a link (junction) from an expected location (c:\hostedtoolcache) to the actual  (D:\hostedtoolcache). 

**Related issue:**
[Original issue](https://github.com/actions/setup-go/issues/240)
[Investigation issue](https://github.com/actions/runner-images-internal/discussions/5143)

**Check list:**

### DIFF
--- a/.github/workflows/windows-validation.yml
+++ b/.github/workflows/windows-validation.yml
@@ -1,4 +1,4 @@
-name: validate Windows installation
+name: Validate Windows installation
 
 on:
   push:
@@ -13,7 +13,7 @@ on:
 jobs:
   create-link-if-not-default:
     runs-on: windows-latest
-    name: 'Setup ${{ matrix.go }}, cache: ${{ matrix.cache }}'
+    name: 'Validate if symlink is created'
     strategy:
       matrix:
         cache: [false, true]
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: non-default-cache-${{ matrix.cache }}-${{ matrix.go }}
+      - name: 'Setup ${{ matrix.cache }}, cache: ${{ matrix.go }}'
         uses: ./
         with:
           go-version: ${{ matrix.go }}
@@ -81,7 +81,7 @@ jobs:
         shell: bash
 
   dont-create-link-if-default:
-    name: 'Use default go, cache: ${{ matrix.cache }}'
+    name: 'Validate if symlink is not created for default go'
     runs-on: windows-latest
     needs: find-default-go
     strategy:
@@ -90,12 +90,13 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: ./
+      - name: 'Setup default go, cache: ${{ matrix.cache }}'
+        uses: ./
         with:
           go-version: ${{ needs.find-default-go.outputs.version }}
           cache: ${{ matrix.cache }}
 
-      - name: 'Drive C: should have Go installation'
+      - name: 'Drive C: should have Go installation, cache: ${{ matrix.cache}}'
         run: |
           size=$(du -m -s 'C:\hostedtoolcache\windows\go\${{ needs.find-default-go.outputs.version }}\x64'|cut -f1 -d$'\t')
           if [ $size -eq 0 ];then
@@ -104,7 +105,7 @@ jobs:
           fi
         shell: bash
 
-      - name: 'Drive D: should not have Go installation'
+      - name: 'Drive D: should not have Go installation, cache: ${{ matrix.cache}}'
         run: |
           if [ -e 'D:\hostedtoolcache\windows\go\${{ needs.find-default-go.outputs.version }}\x64' ];then
             echo 'D:\hostedtoolcache\windows\go\${{ needs.find-default-go.outputs.version }}\x64 should not exist for hosted version of go';
--- a/.github/workflows/windows-validation.yml
+++ b/.github/workflows/windows-validation.yml
@@ -29,7 +29,7 @@ jobs:
       - run: |
           du -m -s 'D:\hostedtoolcache\windows\go\${{ matrix.go }}\x64'
           size=$(du -m -s 'D:\hostedtoolcache\windows\go\${{ matrix.go }}\x64'|cut -f1 -d$'\t')
-          # make sure archive does not take lost of the space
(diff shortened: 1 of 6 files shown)


ilia-shipitsin (approved):

src/installer.ts
```diff
@@ -185,15 +201,44 @@ async function installGoVersion(
...
-  return cachedDir;
+  fs.mkdirSync(path.dirname(defaultToolCacheDir), {recursive: true});
+  fs.symlinkSync(actualToolCacheDir, defaultToolCacheDir, 'junction');
+  core.info(`Created link ${defaultToolCacheDir} => ${actualToolCacheDir}`);
+
+  // restore toolcache root to default drive c:
+  process.env['RUNNER_TOOL_CACHE'] = defaultToolCacheRoot;
+
+  // make outer code to continue using toolcache as if it were installed on c:
+  return defaultToolCacheDir;
```
IvanZosimov, line 241: I'd suggest to incapsulate this code into the wrapper function, which inside of it will decide how to save go into the toolcache.

src/installer.ts
```diff
@@ -164,6 +164,22 @@ async function resolveVersionFromManifest(
...
+  const cachedDir = await tc.cacheDir(
+    extPath,
+    'go',
+    makeSemver(info.resolvedVersion),
+    arch
+  );
+  core.info(`Successfully cached go to ${cachedDir}`);
+  return cachedDir;
+}
+
```
IvanZosimov, line 182: I'd return calls of core.info() into the installGoVersion() function and not wrap tc.cacheDir() into additional function as it's pretty small and reusable piece of code by itself.

src/installer.ts
```diff
@@ -164,6 +164,62 @@ async function resolveVersionFromManifest(
...
+  extPath: string,
+  tool: string,
+  version: string,
+  arch: string
+): Promise<string | false> {
+  if (os.platform() !== 'win32') return false;
+
+  const isHosted =
+    process.env['RUNNER_ENVIRONMENT'] === 'github-hosted' ||
+    process.env['AGENT_ISSELFHOSTED'] === '0';
```
dusan-trickovic, line 179: This is more just a clarifying question, but should this return a bool value?

dusan-trickovic (commented):
Overall looks good to me, I just had to ask a clarifying question before officially approving it. If I made a mistake, feel free to let me know and an approval will come right away.

src/installer.ts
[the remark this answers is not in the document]
MaksimZhukov: I agree with @dusan-trickovic! In my opinion, the current implementation degrades readability and complicates working with the code.

.github/workflows/windows-validation.yml
```diff
@@ -0,0 +1,113 @@
...
+    runs-on: windows-latest
+    name: 'Setup ${{ matrix.go }}, cache: ${{ matrix.cache }}'
+    strategy:
+      matrix:
+        cache: [false, true]
+        go: [1.20.1]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: non-default-cache-${{ matrix.cache }}-${{ matrix.go }}
```
MaksimZhukov, line 24:
```suggestion
      - name: 'Setup ${{ matrix.go }}, cache: ${{ matrix.cache }}'
```

.github/workflows/windows-validation.yml
```diff
@@ -0,0 +1,113 @@
...
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+
+jobs:
+  create-link-if-not-default:
+    runs-on: windows-latest
+    name: 'Setup ${{ matrix.go }}, cache: ${{ matrix.cache }}'
```
MaksimZhukov, line 16:
```suggestion
    name: 'Validate if symlink is created'
```

.github/workflows/windows-validation.yml
```diff
@@ -0,0 +1,113 @@
...
+    steps:
+      - run: |
+          version=`go env GOVERSION|sed s/^go//`
+          echo "default go version: $version"
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+        id: goversion
+        shell: bash
+
+  dont-create-link-if-default:
+    name: 'Use default go, cache: ${{ matrix.cache }}'
```
MaksimZhukov, line 84:
```suggestion
    name: 'Validate if symlink is not created for default go'
```

.github/workflows/windows-validation.yml
```diff
@@ -0,0 +1,113 @@
...
+    runs-on: windows-latest
+    needs: find-default-go
+    strategy:
+      matrix:
+        cache: [false, true]
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: ./
```
MaksimZhukov, line 93:
```suggestion
      - name: 'Setup default go, cache: ${{ matrix.cache }}'
        uses: ./
```

src/installer.ts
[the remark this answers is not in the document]
dsame: process.env['RUNNER_ENVIRONMENT'] === 'github-hosted' - confirms the action runs in github environment 
process.env['AGENT_ISSELFHOSTED'] === '0'  - confirm the action runs in AzDev environment

either condition means we should care about specific hardware configuration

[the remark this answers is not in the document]
MaksimZhukov: Oops... Looks like Dusan's comment is about other things. I was talking about the function in general. Returning a string and boolean is not a good practice in my opinion.

.github/workflows/windows-validation.yml
```diff
@@ -0,0 +1,113 @@
+name: validate Windows installation
```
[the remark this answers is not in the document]
dsame: fixed

src/installer.ts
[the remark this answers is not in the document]
dsame: Why not? It can not return any boolean value but only `false` and it is fully compatible with returning string but it lets avoid implicit "string to boolean" conversion when checking the result and does not require additional convention "empty string stands for non-executed operation"

dmitry-shibanov (approved):

marko-zivic-93 (approved):

Issue #316, opened 2023-01-10
Add option to restore cache only


ForsakenHarmony:
**Description:**
It would be nice to have the option to restore only and not save.

**Justification:**
In bigger repos, it's easy to hit the 10gb cache limit. Not caching in PRs can help with this.
Right now, I have to manually delete caches to help keep the caches on `main` alive.

**Are you willing to submit a PR?**

**References**

https://github.com/actions/toolkit/issues/1308
https://github.com/Swatinem/rust-cache/issues/95 (has a `save-if` option https://github.com/Swatinem/rust-cache#example-usage)

e-korolevskii:
Hello @ForsakenHarmony! Thank you for the suggested idea!
We will consider adding this feature and will let you know as soon as we have any decision.

gwatts:
@ForsakenHarmony likely seen this already, but if not you might find this helpful too to automatically delete PR caches when the PR is closed: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries

ForsakenHarmony:
@gwatts yes, I'm doing that, doing it after each commit's CI is done actually, but it's an ugly workaround

dsame:
Hello @ForsakenHarmony, if i understood your request correctly you need the "restore cache only" action https://github.com/actions/cache/blob/main/restore/README.md

Does it solve the problem?

lzap:
We would love this feature too. Our CI pipeline consist of:

* build job (runs `go build`)
* test job (runs `go test`)

When cache is enabled (default in v4) and there is a change in `go.sum` both jobs run in parallel but because our build job is slower, test job wins and uploads the cache. Problem is, the test job does not build all code path - only those which are executed by unit tests. The next time cache is downloaded, build job will use the cache created by test job which is not 100% so this is leading to wasted CPU cycles.

We would like to configure test job to "download only", basically ability to disable the setup-go post hook uploading the cache.

lzap:
How do I use `lookup-only` with `setup-go`?

dsame:
Hello @lzap,
you need to disable go caching and use https://github.com/actions/cache/blob/main/restore/README.md

lzap:
[...]

PR #400, opened 2023-07-25
Add cache-restore-only input

dsame:
**Description:**

The [suggested workaround](https://github.com/actions/setup-go/issues/316#issuecomment-1632080974) does not really work because it requires the users [to re-engineer and reconstruct](https://github.com/actions/setup-go/issues/316#issuecomment-1632133843) the automatically created cache key 

To avoid it i introduced new input `cache-restore-only` that makes build to skip caching without making user to deal with cache key

**Related issue:**
[link to the related issue.](https://github.com/actions/setup-go/issues/316)

**Check list:**


freeformz:
Yes please. I understand the arguments [against](https://github.com/actions/setup-go/issues/316#issuecomment-1632080974), but respectfully disagree - it's a pita to have to reverse engineer the key setup and copy it into a ton of different repo workflows. And then there is the possibility of future changes. I'd like the setup-go action to "just handle all of that".

gregwebs:
Can this PR get reviewed? A use case I have for this is to be able to speed up deployments. Currently waiting for the go cache cleanup to push at the end adds a noticeable delay and the deployment build is not going to produce a useful cache that doesn't already exist. I am looking to see if it is spending time caching unnecessary things and if there are settings for that, but would be nice to avoid altogether in this case.

devnev:
This would be great to get merged. @HarithaVattikuti ?

bconway:
This would be a great (and simple) addition.

PR #411, opened 2023-08-12
Fix hosted tool cache usage on windows

galargh:
**Description:**
This change ensures that we symlink not only the tool directory but also the completion marker when adding `go` to the hosted tool cache on Windows.

**Related issue:**
Not raised, but without this change it is impossible to run the action twice with the same `go-version` input on Windows because the second run fails with errors related to files already existing.

**Testing:**
- 🔴 Example test run without the fix: https://github.com/galargh/setup-go/actions/runs/5843165063/job/15845040764
- 🟢 Example test run with the fix: https://github.com/galargh/setup-go/actions/runs/5843616834/job/15845918944

**Check list:**
- [x] Mark if documentation changes are required. _NOT REQUIRED_

### DIFF
--- a/.github/workflows/windows-validation.yml
+++ b/.github/workflows/windows-validation.yml
@@ -114,7 +114,7 @@ jobs:
         shell: bash
 
   hostedtoolcache:
-    name: 'Validate if hostedtoolcache works as expected' 
+    name: 'Validate if hostedtoolcache works as expected'
     runs-on: windows-latest
     strategy:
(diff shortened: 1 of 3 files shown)


dmitry-shibanov (approved):

dsame (approved):
I confirm changing "junction" to "dir" does not solve the problem and this fix must be applied

marko-zivic-93 (approved):

Issue #371, opened 2023-04-28
feat: add support for multi-module repos


dolmen:
**Description:**
Add support for caching dependencies for repositories made of multiple Go modules.

Example:
* Repo [dolmen-go/mylogin-driver](https://github.com/dolmen-go/mylogin-driver/) has Go modules [`github.com/dolmen-go/mylogin-driver`](https://github.com/dolmen-go/mylogin-driver/blob/master/go.mod) and [`github.com/dolmen-go/mylogin-driver/cmd/mylogin-query`](https://github.com/dolmen-go/mylogin-driver/blob/master/cmd/mylogin-query/go.mod)

**Justification:**
Each Go module has its own set of dependencies. Dependencies caching should cover the union of the dependencies of all Go modules of the repo.

From the [documentation](https://github.com/actions/setup-go/blob/dd84a9531a6f8e72c321f2aa3b9048ed359670e4/README.md#caching-dependency-files-and-build-outputs), it looks like a single Go module is supported.

**Are you willing to submit a PR?**
No. I'm not familiar with GitHub Actions development.

dsame:
Hello @dolmen, thanks for the suggestion, we are starting to investigate it

dolmen:
Design idea: allow to path an array of path for the `cache-dependency-path` option as an alternative to a string.

dsame:
This feature resembles https://github.com/actions/setup-node/pull/735

dsame:
Hello @dolmen ,

`cache-dependency-path` accepts both list and glob syntax:

```
          cache-dependency-path: |
             hello/go.sum 
             greeting/go.sum
```

```
          cache-dependency-path: "**/*.sum"
```

Sample workflow: https://github.com/akv-demo/test-setup-go/blob/cache-multi/.github/workflows/blank.yml
Sample build: https://github.com/akv-demo/test-setup-go/actions/runs/5799499947/job/15719631207

Did this answer help?

dolmen:
Not yet tested, but it seems to fit my need.

However, what about making `**/go.sum` the default?

dsame:
@dolmen i do not think it is good idea to allow implicit dependencies. I believe the user of the action should have clear understanding of where the dependencies come from.

dsame:
@dolmen i am going to close the issue because the problem is resolved without the implementing requested feature, but please feel free to reopen it or create new one in case if the problem still exists

artemgavrilov:
[...]

PR #417, opened 2023-08-25
 Improve documentation regarding dependencies caching

artemgavrilov:
**Description:**
Just a highlighting that `cache-dependency-path` parameter can accept multiple files and can handle wildcards + example.

**Related issue:**
https://github.com/actions/setup-go/issues/371

**Check list:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ The `cache` input is optional, and caching is turned on by default.
 
 The action defaults to search for the dependency file - go.sum in the repository root, and uses its hash as a part of
 the cache key. Use `cache-dependency-path` input for cases when multiple dependency files are used, or they are located
-in different subdirectories. Wildcards also supported.
+in different subdirectories. Globs also supported.
 
 If some problem that prevents success caching happens then the action issues the warning in the log and continues the execution of the pipeline. 
 
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ The `cache` input is optional, and caching is turned on by default.
 
 The action defaults to search for the dependency file - go.sum in the repository root, and uses its hash as a part of
 the cache key. Use `cache-dependency-path` input for cases when multiple dependency files are used, or they are located
-in different subdirectories. Globs also supported.
+in different subdirectories. The input supports glob patterns.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -159,7 +159,7 @@ The `cache` input is optional, and caching is turned on by default.
 
 The action defaults to search for the dependency file - go.sum in the repository root, and uses its hash as a part of
 the cache key. Use `cache-dependency-path` input for cases when multiple dependency files are used, or they are located
-in different subdirectories.
+in different subdirectories. Wildcards also supported.
```
dsame, line 162: To precise this syntax is not "wildcards" but rather "glob pattern" (https://en.wikipedia.org/wiki/Glob_(programming)). The difference is significant because the latter much more powerful and calling it "wildcards" could be misleading.

artemgavrilov: Fixed.

marko-zivic-93 (approved):

README.md
```diff
@@ -159,7 +159,7 @@ The `cache` input is optional, and caching is turned on by default.
 
 The action defaults to search for the dependency file - go.sum in the repository root, and uses its hash as a part of
 the cache key. Use `cache-dependency-path` input for cases when multiple dependency files are used, or they are located
-in different subdirectories.
+in different subdirectories. Globs also supported.
```
IvanZosimov, line 162:
```suggestion
in different subdirectories. The input supports glob patterns.
```
I'd expand the sentence.

dsame (approved):

README.md
```diff
@@ -172,7 +172,11 @@ steps:
     with:
       go-version: '1.17'
       check-latest: true
-      cache-dependency-path: subdir/go.sum
+      cache-dependency-path: |
+             subdir/go.sum 
+             tools/go.sum
```
yurishkuro, line 177: this syntax is pretty unclear. Is this parameter accepting space-separated strings? Or an array? Something like this would be more idiomatic yaml:
```
      cache-dependency-path:
        - subdir/go.sum 
        - tools/go.sum
```

artemgavrilov: @yurishkuro This format inherited from [cache](https://github.com/actions/cache) action as I understood: https://github.com/actions/cache#using-a-combination-of-restore-and-save-actions

So I think it make sense to open an issue there.

PR #422, opened 2023-09-13
Problem matchers: Matches of test failures

WillAbides:
**Description:**

Currently test failures do not get annotated because the problem matcher requires that a column number be present, but go's test output does not include a column for test failures.

An example line that gets missed by the current matcher is:
```
/path/to/main_test.go:13: expected true but got false
```

The updated matcher makes the column number optional and matches the line above.

**Check list:**


dsame:
Hello @WillAbides , we are starting to review the PR

dsame (approved):

marko-zivic-93 (approved):

WillAbides:
You might want to hold off on this one for a bit. I used this same strategy in another action and got [this report](https://github.com/WillAbides/setup-go-faster/issues/27) about it.

This pattern also matches test log statements, so those are getting annotated as failures even when the test passes.

PR #426, opened 2023-09-18
PoC PR for next-gen caching

dsame:
**Description:**

The existing set of cache related inputs ...
```
  cache:
    description: Used to specify whether caching is needed. Set to true, if you'd like to enable caching for both modules and intermediate build files.
    default: true
  cache-dependency-path:
    description: 'Used to specify the path or glob pattern to a dependency file(s), default: go.sum'
```

... is expanded by the additional ones:

```
  cache-mod:
    description: Used to specify whether modules caching is needed. Set to false, if you've found it increase the overall build time.
    default: true
  cache-build:
    description: Used to specify whether caching of intermediate build files is needed. Set to false, if you've found it increase the overall build time.
    default: true
  cache-build-path:
    description: 'Used to specify the path or glob pattern to source files(s) that affect the build cache, default: **/*.go'
  cache-id:
    description: 'Used to modify cache ID if the parallel workflows must not share the same cache, default: none'
  cache-lookup-only:
    description: 'Use the cache created by another workflow, but do not update it'
    default: false
```

Besides, caches of modules and intermediate build artifacts are kept and updated separately.

**Related issue:**
 https://github.com/actions/setup-actions-team/issues/39

**Check list:**

### DIFF
--- a/__tests__/cache-restore.test.ts
+++ b/__tests__/cache-restore.test.ts
@@ -41,7 +41,7 @@ describe('restoreCache', () => {
 
     //Act + Assert
     await expect(async () => {
-      await cacheRestore.restoreCache(
+      await cacheRestore.restoreModCache(
         versionSpec,
         packageManager,
         cacheDependencyPath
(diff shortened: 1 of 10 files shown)


irl-segfault:
These would be very convenient and useful knobs to expose.  Thanks for pushing this towards merging!

Issue #424, opened 2023-09-15
Tar errors on cache restore after toolchain installation


matthewhughes-uw:
**Description:**

When the version of Go in the `go` directive of `go.mod` is different from the version in the `toolchain` directive there is a cache conflict which causes errors on cache extraction.

Note: I think this is related to, but different from https://github.com/actions/setup-go/issues/403 since that appears to  be covering the situation where some other tool is also working on the cache, but this issue is purely for the action running on a hosted runner without any additional changes to the environment etc.

**Action version:**

`actions/setup-go@v4`

**Platform:**
- [x] Ubuntu
- [ ] macOS (untested)
- [ ] Windows (untested)

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**

* `go 1.21.0`
* `toolchain go1.21.1`

**Repro steps:**  

See repo https://github.com/matthewhughes-uw/setup-go-test/, very basic repo with a single workflow that just runs this action (https://github.com/matthewhughes-uw/setup-go-test/blob/main/.github/workflows/go.yaml):

First run of the action https://github.com/matthewhughes-uw/setup-go-test/actions/runs/6195831453/job/16821285537

* Installs Go 1.21.0 (version detected from `go-version-file`
* Go detects toolchain and performs a toolchain install of Go 1.21.1 (see output `go: downloading go1.21.1 (linux/amd64)`)
* This places some files under `../../../go/pkg/mod/golang.org/toolchain@v0.0.1-go1.21.1.linux-amd64`
* Finds no existing cache, doesn't try to extract anything
* Caches  `/home/runner/go/pkg/mod` (which contains the files above) and `/home/runner/.cache/go-build`

Second run of the action https://github.com/matthewhughes-uw/setup-go-test/actions/runs/6195844389/job/16821321968

* First two steps as above
* Finds an existing cache, attempts to extract it, but because of the toolchain installation step files under `../../../go/pkg/mod/golang.org/toolchain@v0.0.1-go1.21.1.linux-amd64` are already present on the machine, causing errors from `tar`

**Expected behavior:**

Caching of the toolchain is handled gracefully, no errors between runs that don't contain any extra caching logic outside of what the action provides by default

**Actual behavior:**

Errors on cache extraction (see repro steps)

dsame:
Hello @matthewhughes-uw , thank you for the input. We are starting to investigate the problem.

thall:
[...]

PR #456, opened 2024-02-08
Avoid Toolchain download before cache download

matthewhughes-uw:
**Description:**

`go version` is run before downloading the cache, but if this is run with a version of `go` that triggers a Toolchain download[1], e.g. if the installed Go is 1.20.0 but `go.mod` has a toolchain directive `go1.20.1` then a toolchain is downloaded to e.g.
`$GOMODCACHE/golang.org/toolchain@v0.0.1-go1.21.1.linux-amd64`, if this file already exists in the cache (e.g. this is the second run of this action we not cache invalidation) then the cache download will try and overwrite these files resulting in noisy errors like:

    /usr/bin/tar: ../../../go/pkg/mod/golang.org/toolchain@v0.0.1-go1.21.6.linux-amd64/lib/time/mkzip.go: Cannot open: File exists

Instead, force `go` to use the local toolchain (i.e. the one the one that shipped with the go command being run) via setting the `GOTOOLCHAIN` environment variable[1]:

> When GOTOOLCHAIN is set to local, the go command always runs the
bundled Go toolchain.

[1] https://go.dev/doc/toolchain#select

**Related issue:**

This addresses https://github.com/actions/setup-go/issues/424

**Check list:**

### DIFF
--- a/dist/cache-save/index.js
+++ b/dist/cache-save/index.js
@@ -58600,8 +58600,8 @@ const cache = __importStar(__nccwpck_require__(7799));
 const core = __importStar(__nccwpck_require__(2186));
 const exec = __importStar(__nccwpck_require__(1514));
 const package_managers_1 = __nccwpck_require__(6663);
-const getCommandOutput = (toolCommand) => __awaiter(void 0, void 0, void 0, function* () {
-    let { stdout, stderr, exitCode } = yield exec.getExecOutput(toolCommand, undefined, { ignoreReturnCode: true });
+const getCommandOutput = (toolCommand, execOpts) => __awaiter(void 0, void 0, void 0, function* () {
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -62,7 +62,10 @@ export async function run() {
     core.debug(`add bin ${added}`);
 
     const goPath = await io.which('go');
-    const goVersion = (cp.execSync(`${goPath} version`) || '').toString();
+    // run `go version` with the bundled Go toolchain to avoid potentially
+    // downloading one into the cache
+    const goVersion = (cp.execSync(`${goPath} version`) || '',
+    {env: {...process.env, GOTOOLCHAIN: 'local'}}).toString();
```
matthewhughes-uw, line 68: Happy to hear any suggestions for testing this

matthewhughes-uw:
https://github.com/actions/setup-go/pull/460 is a better solution to this

Issue #457, opened 2024-02-14
More specific handling/detection of Go toolchain versions


matthewhughes934:
**Description:**

The goal is to just have an approach to manage toolchain installs, the following is a suggestion and open to discussion/changes:

* If the user provides a specific version, ensure we use exactly that version (possibly requires setting `GOTOOLCHAIN=local` in the environment, see [docs](https://go.dev/doc/toolchain#select))
* If the user provides a `go-version-file`, and:
    * The file has a `toolchain` directive: then use that since the [docs say](https://go.dev/doc/toolchain#config)
       > The toolchain line declares a suggested toolchain to use with the module or workspace

       This is in comparison with the `go` version, which is the minimum version of Go required to install the module (the assumption being: we're working within the module, not install it)
    * The file has no `toolchain` directive: then use the version from the `go` directive (current behaviour)

**Justification:**

At the moment this action does not consider [toolchain versions](https://go.dev/doc/toolchain). This leads to:

* Some duplicate work, the action will install a version of Go, a toolchain will be detected, the toolchain will be detected and then _another_ version of Go installed
* Unexpected behaviour: if you specify this action runs with some Go version (e.g. `1.21.0`) but your `go.mod` contains a toolchain directive for a newer version (e.g. `1.22.0`) then, without any other configuration/environment setup, any `go` commands will be run using `go 1.22.0`
* https://github.com/actions/setup-go/issues/424 is related to not managing toolchain installs

You can see this in this [example repo](https://github.com/matthewhughes934/setup-go-toolchain-examples), which is just a  repo with a `go.mod` whose `go` directive is `1.21.0` and whose toolchain directive is `1.22.0` and some actions that display the version of Go under some conditions, of interest is when we [specify the action with version `1.21.0`](https://github.com/matthewhughes934/setup-go-toolchain-examples/blob/main/.github/workflows/go.yaml#L6): we ask for `1.21.0` but a toolchain install will result in `1.22.0` being installed and run

**Are you willing to submit a PR?**

Yep, but there should be some discussion around which approach to take before implementing anything

EDIT: I noticed that the official Go docker images also set `GOTOOLCHAIN=local` https://github.com/docker-library/golang/blob/dae3405a325073e8ad7c8c378ebdf2540d8565c4/Dockerfile-linux.template#L162C3-L163C1 with some discussion on the linked issue there: https://github.com/docker-library/golang/issues/472

matthewhughes934:
[...]

PR #460, opened 2024-03-03
Improve toolchain handling

matthewhughes934:
- Configure environment to avoid toolchain installs

    Force `go` to always use the local toolchain (i.e. the one the one that
    shipped with the go command being run) via setting the `GOTOOLCHAIN`
    environment variable to `local`[1]:

    > When GOTOOLCHAIN is set to local, the go command always runs the
    bundled Go toolchain.

    This is how things are setup in the official Docker images (e.g.[2], see
    also the discussion around that change[3]). The motivation behind this
    is to:

    * Reduce duplicate work, the action will install a version of Go, a
      toolchain will be detected, the toolchain will be detected and then
      another version of Go installed[4]
    * Avoid Unexpected behaviour: if you specify this action runs with some Go
      version (e.g. `1.21.0`) but your go.mod contains a `toolchain` or `go`
      directive for a newer version (e.g. `1.22.0`) then, without any other
      configuration/environment setup, any go commands will be run using go
      `1.22.0`
    * TODO: link image

    This will be a **breaking change** for some workflows. Given a `go.mod`
    like:

        module proj

        go 1.22.0

    Then running any `go` command, e.g. `go mod tidy`, in an environment
    where only go versions before `1.22.0` were installed would previously
    trigger a toolchain download of Go `1.22.0` and that version being used
    to execute the command. With this change the above would error out with
    something like:

    > go: go.mod requires go >= 1.22.0 (running go 1.21.7;
    GOTOOLCHAIN=local)

    Link:  https://go.dev/doc/toolchain#select [1]
    Link:  https://github.com/docker-library/golang/blob/dae3405a325073e8ad7c8c378ebdf2540d8565c4/Dockerfile-linux.template#L163 [2]
    Link:  https://github.com/docker-library/golang/issues/472 [3]
    Link:  https://github.com/actions/setup-go/issues/424 [4]
    Issue: https://github.com/actions/setup-go/issues/457

- Prefer installing version from `toolchain` directive

    Prefer this over the version from the `go` directive. Per the docs[1]

    > The toolchain line declares a suggested toolchain to use with the
    module or workspace

    It seems reasonable to use this, since running this action in a
    directory containing a `go.mod` (or `go.work`) suggests the user is
    wishing to work _with the module or workspace_.

    Link: https://go.dev/doc/toolchain#config [1]
    Issue: https://github.com/actions/setup-go/issues/457

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -94312,6 +94312,7 @@ var __importDefault = (this && this.__importDefault) || function (mod) {
     return (mod && mod.__esModule) ? mod : { "default": mod };
 };
 Object.defineProperty(exports, "__esModule", ({ value: true }));
+exports.GOTOOLCHAIN_LOCAL_VAL = exports.GOTOOLCHAIN_ENV_VAR = void 0;
 exports.getGo = getGo;
 exports.extractGoArchive = extractGoArchive;
 exports.getManifest = getManifest;
@@ -94330,6 +94331,8 @@ const sys = __importStar(__nccwpck_require__(5632));
 const fs_1 = __importDefault(__nccwpck_require__(7147));
 const os_1 = __importDefault(__nccwpck_require__(2037));
 const utils_1 = __nccwpck_require__(1314);
+exports.GOTOOLCHAIN_ENV_VAR = 'GOTOOLCHAIN';
+exports.GOTOOLCHAIN_LOCAL_VAL = 'local';
 const MANIFEST_REPO_OWNER = 'actions';
 const MANIFEST_REPO_NAME = 'go-versions';
 const MANIFEST_REPO_BRANCH = 'main';
@@ -94783,12 +94786,12 @@ const os_1 = __importDefault(__nccwpck_require__(2037));
 function run() {
     return __awaiter(this, void 0, void 0, function* () {
         try {
-            setToolchain();
             //
             // versionSpec is optional.  If supplied, install / use from the tool cache
             // If not supplied then problem matchers will still be setup.  Useful for self-hosted.
             //
             const versionSpec = resolveVersionInput();
+            setGoToolchain();
             const cache = core.getBooleanInput('cache');
             core.info(`Setup go version spec ${versionSpec}`);
             let arch = core.getInput('architecture');
@@ -94897,20 +94900,18 @@ function resolveVersionInput() {
     }
     return version;
 }
-function setToolchain() {
+function setGoToolchain() {
     // docs: https://go.dev/doc/toolchain
     // "local indicates the bundled Go toolchain (the one that shipped with the go command being run)"
     // this is so any 'go' command is run with the selected Go version
     // and doesn't trigger a toolchain download and run commands with that
     // see e.g. issue #424
-    // and a similar discussion: https://github.com/docker-library/golang/issues/472
-    const toolchain = 'local';
-    const toolchainVar = 'GOTOOLCHAIN';
-    // set the value in process env so any `go` commands run as child-process
+    // and a similar discussion: https://github.com/docker-library/golang/issues/472.
+    // Set the value in process env so any `go` commands run as child-process
     // don't cause toolchain downloads
-    process.env[toolchainVar] = toolchain;
+    process.env[installer.GOTOOLCHAIN_ENV_VAR] = installer.GOTOOLCHAIN_LOCAL_VAL;
     // and in the runner env so e.g. a user running `go mod tidy` won't cause it
-    core.exportVariable(toolchainVar, toolchain);
+    core.exportVariable(installer.GOTOOLCHAIN_ENV_VAR, installer.GOTOOLCHAIN_LOCAL_VAL);
 }
 
 
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -94777,6 +94777,7 @@ const os_1 = __importDefault(__nccwpck_require__(2037));
 function run() {
     return __awaiter(this, void 0, void 0, function* () {
         try {
+            setToolchain();
             //
             // versionSpec is optional.  If supplied, install / use from the tool cache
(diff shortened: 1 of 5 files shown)


matthewhughes934:
Testing this action https://github.com/matthewhughes934/setup-go-test, see the workflow runs for details https://github.com/matthewhughes934/setup-go-test/actions

reneleonhardt:
Did you rebase already? GitHub doesn't allow me to see the parent commit.
npm packages are very old, maybe `npm audit --fix` will help after rebasing.

matthewhughes934:
> Did you rebase already? GitHub doesn't allow me to see the parent commit. npm packages are very old, maybe `npm audit --fix` will help after rebasing.

The vulnerability reported is also present on `main`:

```console
$ git checkout main
$ git rev-parse HEAD
8e57b58e57be52ac95949151e2777ffda8501267
$ npm audit --audit-level=high
# npm audit report

form-data  >=4.0.0 <4.0.4 || <2.5.4
Severity: critical
form-data uses unsafe random function in form-data for choosing boundary - https://github.com/advisories/GHSA-fjxv-7rqg-78g4
form-data uses unsafe random function in form-data for choosing boundary - https://github.com/advisories/GHSA-fjxv-7rqg-78g4
fix available via `npm audit fix`
node_modules/@azure/core-http/node_modules/form-data
node_modules/@types/node-fetch/node_modules/form-data
node_modules/form-data

1 critical severity vulnerability

To address all issues, run:
  npm audit fix
```

The vulnerability is two days old: https://github.com/advisories/GHSA-fjxv-7rqg-78g4, here's a separate PR for that https://github.com/actions/setup-go/pull/618 (though I'm not sure why `dependabot` hasn't raised one?)

reneleonhardt:
Dependabot doesn't always work as you would expect, it's not resilient for example, a simple network error can disable updates. Here is your failed run from 2 days ago, only maintainers are allowed to read the logs:
https://github.com/actions/setup-go/actions/runs/16426449639
It doesn't even store state between runs, and when you only allow it to run once per week like in this repo, then it could take days or weeks until you finally could merge even one PR again...

As long as AI reviews haven't been enabled, only manual maintainer work could speed-up reviews.
After waiting for one year for a few lines of code, I wouldn't hold my breath...

matthewhughes934:
I've dropped the commit that changed behaviour from install the Go version specified in the `toolchain` directive to the one specified in the `go` directive, since I _think_ this is what people expect this action to do (it's how things worked until `toolchains` were introduced to the language). Maybe worth a separate option to install the `toolchain` version.

kaovilai:
I'm holding my breath based on https://github.com/actions/setup-go/pull/618

lmvysakh:
Hi @matthewhughes934,

Thank you for your thoughtful work on this PR, After reviewing the code changes in detail, We wanted to highlight a specific scenario where the current implementation does not fully satisfy one of our core criteria:

**Scenario:** When a workflow uses only a go.mod file that contains both a go directive and a toolchain directive (with the toolchain version being higher), and there is no explicit go-version or GOTOOLCHAIN=local set in the workflow.

**Current Code Behavior:** The action installs and uses the Go version from the go directive, rather than the one specified in the toolchain directive.

**Expected Behavior (per our criteria and official Go documentation):** In this situation, the action should detect the presence of the toolchain directive in go.mod and install or use the toolchain version (for example, go1.22.6). The toolchain directive is meant to indicate the intended Go toolchain for the project and should take precedence if no explicit workflow overrides are present.

- The current code sets GOTOOLCHAIN=local at startup, which works well for scenarios with explicit workflow versioning or environment overrides.
- However, when neither is provided, the logic still defaults to the go directive, missing the opportunity to honor the toolchain directive as intended.

Could you please update the logic so that when only a go.mod file is present and both directives exist, the action installs and uses the Go version from the toolchain directive unless overridden by the workflow or environment? You can also make use of implementing the go-version-directive option if required, which you have suggested as per your earlier comment. This adjustment will ensure that all scenarios are covered and that the action's behavior aligns with both our criteria and the expectations of Go users relying on toolchain directives.

Additionally, please update the documentation to reflect any changes or new options you introduce, so users are aware of how the action now handles these scenarios and how to configure it for their needs. Once this is addressed, we can discuss progressing the PR further. Thank you again for your contribution !

dist/setup/index.js
```diff
@@ -94663,8 +94663,14 @@ function parseGoVersionFile(versionFilePath) {
     const contents = fs_1.default.readFileSync(versionFilePath).toString();
     if (path.basename(versionFilePath) === 'go.mod' ||
         path.basename(versionFilePath) === 'go.work') {
-        const match = contents.match(/^go (\d+(\.\d+)*)/m);
-        return match ? match[1] : '';
+        // toolchain directive: https://go.dev/ref/mod#go-mod-file-toolchain
+        const matchToolchain = contents.match(/^toolchain go(\d+(\.\d+)*)/m);
```
reneleonhardt, line 94667: That regex matches
```
toolchain go1       # invalid (major isn't allowed)
toolchain go1.2.3.4 # invalid (not minor or semver)
```
but not
```
toolchain go1.25rc1 # valid (prerelease)
```
https://go.dev/doc/toolchain

Better:
```suggestion
        const matchToolchain = contents.match(^toolchain go(1\.\d+(?:\.\d+|rc\d+)?));
```

matthewhughes934: > That regex matches
> 
> ```
> toolchain go1       # invalid (major isn't allowed)
> toolchain go1.2.3.4 # invalid (not minor or semver)
> ```
> 
> but not
> 
> ```
> toolchain go1.25rc1 # valid (prerelease)
> ```
> 
> https://go.dev/doc/toolchain
> 
> Better:

See b967a467e1c904489be015f76916b6bb83d60ef2, the check on the `go` directive should probably use the same pattern, but I think that can be a separate change

HarithaVattikuti (approved):

PR #493, opened 2024-07-25
fix: add arch to cache key

Zxilly:
**Description:**

GitHub has added the arm64 runner, so arch should be used as part of the cache key. Otherwise, if there are `macos-13` `macos-14` running at the same time, the late executor will get the wrong cache.

**Related issue:**

https://github.com/actions/setup-python/pull/896

**Check list:**


Zxilly:
@kolyshkin This mainly happens on macOs amd64/arm64, and the previous patch didn't resolve it.

aparnajyothi-y (approved):

aparnajyothi-y (approved):

kolyshkin:
> @kolyshkin This mainly happens on macOS amd64/arm64, and the previous patch didn't resolve it.

I agree. Yet one day we'll have Windows and Linux on ARM and this will be needed there as well. Thanks for adding, LGTM(NB).

Zxilly:
For GitHub Enterprise we already have Linux arm64 provided by ARM.

priyagupta108 (approved):

HarithaVattikuti (approved):

Issue #569, opened 2025-03-11
`go-build` directory should still be cached if a cache hit occured.


kminehart:
**Description:**
the Go build cache, among many things, stores the test cache. I propose that, even in the event of a cache hit, we save the `go-build` directory so that the new test results and build cache can be saved.

**Justification:**
If the test cache does not get saved, then over a short period time, the cache that is saved will not be applicable. Saving the build cache even when there was a cache hit will result in less repeat test runs.

This could result in a much larger cache being stored. Maybe this should require a flag, like `cache-on-failure`?

**Are you willing to submit a PR?**

Yes!

mahabaleshwars:
Hi @kminehart,
Thank you for creating this feature request. We will investigate it and provide feedback as soon as we have some updates.

kminehart:
I realized now that this isn't possible to safely do with GitHub actions cache. depot.dev accomplishes this with [`GOCACHEPROG`](https://depot.dev/blog/go-remote-cache).

PR #570, opened 2025-03-11
Add the 'always-cache' option

kminehart:
**Description:**
The `always-cache` option will always update the cache, even in the event of a cache hit. This is particularly useful for larger projects with a lot of unit tests and a high volume of changes.

**Related issue:**
https://github.com/actions/setup-go/issues/569

**Check list:**


kminehart:
Ah. I just tested this and realized that this won't work; it will fail to save the cache with a key that already exists, and `@actions/cache` has no function to delete or overwrite a cache key.

PR #623, opened 2025-07-30
Fix line-endings in `package.json`

matthewhughes934:
Because I just spent far too much time trying to figure out why `npm install` added carriage returns to `package-lock.json` and then `git` would complain:

> warning: in the working copy of 'package-lock.json', CRLF will be
replaced by LF the next time Git touches it

This is simply the result of running:

    git add --renormalize package.json

Since the attributes on that file set line endings to simply `\n`:

    $ git check-attr -a package.json
    package.json: text: auto
    package.json: eol: lf


v-priyagupta108:
Thanks for raising this.

`package.json` has since been normalized to LF as part of #789, which is now merged, so this change is already in `main`. Closing as superseded.

PR #665, opened 2025-10-17
Fall back to downloading from go.dev/dl instead of storage.googleapis.com/golang

### LINKED ISSUE #664: storage.googleapis.com/golang no longer allows anonymous access causing 403 on Golang download fallback
**Description:**
`actions/setup-go` falls back to downloading from `storage.googleapis.com/golang` when the download from `golang-versions` does not succeed.

Anonymous access to this bucket is no longer allowed, causing 403s:

https://storage.googleapis.com/golang/go1.25.3.linux-amd64.tar.gz

Image

**Action version:**
`actions/setup-go@v4`

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [ ] Hosted
- [x] Self-hosted

**Tools version:**
N/A

**Repro steps:**  
1. Hit the rate limit to `go-versions`. Fall back to downloading from `storage.googleapis.com/golang`.
2. Receive 403

**Expected behavior:**
The fallback to download from Go directly should work.

**Actual behavior:**

```
API rate limit exceeded for <ip>. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
Falling back to download directly from Go
Install from dist
Acquiring go1.25.3 from https://storage.googleapis.com/golang/go1.25.3.linux-amd64.tar.gz
Error: Failed to download version ^1.20.0: Error: Unexpected HTTP response: 403
```

nicholasngai:
**Description:**

storage.googleapis.com/golang no longer allows anonymous bucket access:

Image

Fall back to downloading from dl.google.com/go, which is where https://go.dev/dl/ points to now.

**Related issue:**

Closes #664.

**Check list:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ See full release notes on the [releases page](https://github.com/actions/setup-g
 The action will first check the local cache for a version match. If a version is not found locally, it will pull it from
 the `main` branch of the [go-versions](https://github.com/actions/go-versions/blob/main/versions-manifest.json)
 repository. On miss or failure, it will fall back to downloading directly
-from [go dist](https://dl.google.com/go). To change the default behavior, please use
+from [go dist](https://go.dev/dl). To change the default behavior, please use
 the [check-latest input](#check-latest-version).
 
 **Note:** The `setup-go` action uses executable binaries which are built by Golang side. The action does not build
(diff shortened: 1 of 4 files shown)


dmitshur:
Consider using https://go.dev/dl/ itself and following the redirects. It's likely that https://dl.google.com/go/ will remain current for a long time, but if it ends up changing in the future, this will break and need updating again. So using https://go.dev/dl/ URLs and following where it redirects should be more future-proof.

nicholasngai:
Makes sense to me! Just wasn’t sure which would be more stable—I could see an argument that go.dev is unstable too since they could move their download page to a different URL slug so I just chose one arbitrarily.

Updated the PR to do go.dev/dl though!

dmitshur:
Thanks for doing that!

From my perspective, changing download URLs from `https://storage.googleapis.com/golang/{file}` to `https://go.dev/dl/{file}` is the right thing to do, so I'll leave further PR review to owners of this repository.

priyagupta108 (approved):

aparnajyothi-y (approved):

tdyas:
Why not use the same method as the `go` tool itself? That is, download from the Go module proxy. I opened https://github.com/actions/setup-go/issues/678 for this idea.

PR #729, opened 2026-03-03
fix: allow multiple invocations with caching enabled

maxcleme:
## Summary

This PR fixes an issue where calling `setup-go` multiple times with caching enabled in the same workflow would fail. The second invocation would attempt to save to the same cache key, causing a conflict.

## Problem

When a workflow needs to call `setup-go` multiple times with different configurations (e.g., different working directories with different `go.mod` files), the cache-save post action would fail because:
1. The first invocation successfully saves to the cache
2. The second invocation tries to save to the same cache key (which already exists)
3. This results in errors or warnings about duplicate cache saves

## Solution

This fix introduces state tracking to prevent duplicate cache save attempts:

### New Constants (`src/constants.ts`)
- `getAlreadyCachedKey()` / `setAlreadyCachedKey()`: Track which cache keys were already found in cache (cache hit)
- `getPrimaryCacheKey()` / `setPrimaryCacheKey()`: Track the primary cache key for each invocation
- `getCachedGoModPath()` / `setCachedGoModPath()`: Track which `go.mod` paths have already been cached

### Cache Restore Changes (`src/cache-restore.ts`)
- Store state about whether cache was hit and the primary key used
- Track the `go.mod` path being processed

### Cache Save Changes (`src/cache-save.ts`)
- Check if this specific `go.mod` path was already cached in a previous invocation
- Skip cache save if already handled, preventing duplicate save attempts
- Log appropriate messages when skipping duplicate saves

## Testing

- Added comprehensive tests for the multiple invocation scenario in `__tests__/cache-restore.test.ts`
- Added new test file `__tests__/cache-save.test.ts` with tests for:
  - Normal cache save behavior
  - Skipping save when cache was already hit
  - Skipping save when path was already cached by previous invocation
  - Handling of different `go.mod` paths

## Example Workflow That Now Works

```yaml
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      
      - uses: actions/setup-go@v5
        with:
          go-version: '1.21'
          cache: true
          cache-dependency-path: 'service-a/go.sum'
      
      - run: go build ./...
        working-directory: service-a
      
      - uses: actions/setup-go@v5
        with:
          go-version: '1.21'
          cache: true
          cache-dependency-path: 'service-b/go.sum'
      
      - run: go build ./...
        working-directory: service-b
```

## Checklist

- [x] Source code changes
- [x] Rebuilt dist files (`npm run build`)
- [x] Added/updated tests
- [x] All tests pass

### DIFF
--- a/__tests__/cache-restore.test.ts
+++ b/__tests__/cache-restore.test.ts
@@ -6,6 +6,7 @@ import fs from 'fs';
 import * as cacheRestore from '../src/cache-restore';
 import * as cacheUtils from '../src/cache-utils';
 import {PackageManagerInfo} from '../src/package-managers';
+import {State} from '../src/constants';
 
 describe('restoreCache', () => {
   let hashFilesSpy: jest.SpyInstance;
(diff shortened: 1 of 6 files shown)


maxcleme:
This is vibe crap.

